PR #2, opened 2017-02-13
Compact server component

DanicaPivalica:


src/components/CompactServer/CompactServer.tsx
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    render() {
+         let isCritical = this.props.status === ServerStatus.Critical;
+         let isWarning = this.props.status === ServerStatus.Warning;
+         let isOK = this.props.status === ServerStatus.OK;
+         let className = 'compact-server-container ' ;
+
+         
+        if (isWarning) {
```
hbagarich: we use classnames module for creating the classnames for html elements

src/components/CompactServer/CompactServer.tsx
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    getRoleDisplay(role) {
+        return role.display;
+    }
+    
+     public shouldComponentUpdate(nextProps: ICompactServerProps, nextState) {
+        return !(this.props.serverName === nextProps.serverName
+            && this.props.status === nextProps.status
+            && this.props.roleList.map(this.getRoleDisplay).sort().join(',') === nextProps.roleList.map(this.getRoleDisplay).sort().join(',')
```
hbagarich: can we make a reference check here as to avoid deep comparing two arrays in should component update?

src/components/CompactServer/CompactServer.tsx
DanicaPivalica: Noted! Will change to using classnames.

src/components/CompactServer/CompactServer.tsx
DanicaPivalica: Yes, that is a better solution 💃 Will change it in the dashboard related branch.

PR #3, opened 2017-02-14
toggle switch added

MarinBozinovic:
Created toggle switch component


src/components/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -0,0 +1,49 @@
...
+            
+            className,
+            'toggle-switch'
+        );
+        const slidersClassName = classNames(
+            'toggle-slider'
+        );
+        return(
+            <label className={switchClassName}>
+                <input type="checkbox" onChange={this._onChange} checked={this.state.checked}/>
```
vinkobedek: you are using state here, but what if props.checked is not undefined? we should use the props value here in such a case. or even better, shouldn't we calculate the state on componentwillreceiveprops so that the state is in sync across the component? There is another component that does something like this. I believe it is the default checkbox component.

PR #4, opened 2017-02-16
Line chart

mpisacic:


src/components/LineChart/LineChart.props.ts
```diff
@@ -0,0 +1,11 @@
+export interface IChartProps {
+    title: string;
+    width: number;
+    height: number;
+    data: IChart[];
+}
+
+export interface IChart {
```
vinkobedek: Could we rename this interface? IChart does not really say that it is related to Series points. IChartSeriesPoint? or something like that perhaps

src/components/LineChart/LineChart.props.ts
```diff
@@ -0,0 +1,11 @@
...
+    time: Date;
```
vinkobedek: what if i wanted to show something other than date on the x axis? could we make this a bit more generic?

PR #9, opened 2017-02-21
Line chart component

mpisacic:


src/components/LineChart/LineChart.props.ts
```diff
@@ -2,10 +2,12 @@ export interface IChartProps {
...
-    data: IChart[];
+    gridSize?: number;
+    data: IChartData[];
+    xAxisScale: string; // 'LINEAR' or 'TIME' 
 }
 
-export interface IChart {
-    time: Date;
+export interface IChartData {
+    key: Date | number;
```
vinkobedek: Personally i would prefer 'argument' and value as names(for all chart types). see: https://documentation.devexpress.com/#CoreLibraries/DevExpressXtraChartsSeriesPointClassPropTopic
yeah, yeah, it's windows forms.... But i believe that the naming is correct.

vinkobedek: what does d3 use as names?

mpisacic: Not sure I think that it doesn't matter what data is called when we pass it in d3 we just need to specify what name of data is for what. I will change name to 'argument' it gives it a better meaning.

PR #10, opened 2017-02-21
Pie chart component

mpisacic:


src/components/PieChart/PieChart.tsx
```diff
@@ -0,0 +1,195 @@
...
+            .text(d.data.text);
+    }
+
+    private calculateClass(d: DataType) {
+        if (d.label === 'free') { return 'disk-usage-free'; }
+        
+        let className;
+
+        if (this.props.statusInterval) {
+            const okInterval = this.props.statusInterval.filter((d) => d.status === 'ok')[0];
```
vinkobedek: this entire region of code is awfully specific. Shouldn't this component be a bit more generic?

mpisacic: We could make component drawing some random (or specified) colors that aren't tied up to classes or anything, so every part of chart will be in some of specified color.

vinkobedek: usually there is a default pallete defined, but you can override it for your "series".

src/components/PieChart/PieChart.tsx
```diff
@@ -0,0 +1,179 @@
+import * as React from 'react';
+import * as d3 from 'd3';
+import * as classNames from 'classnames';
+import { Label } from '../Label/Label';
+import { IPieChartProps } from './PieChart.props';
+
+export type DataType = {
```
hbagarich: can this be replaced with IPieChartData interface?

src/components/PieChart/PieChart.tsx
mpisacic: Added default pallette and classes so color can be redefined in container component above this one.

src/components/PieChart/PieChart.tsx
mpisacic: Did you mean replacing DataType with IPieChartData interface?

src/components/PieChart/PieChart.tsx
```diff
@@ -170,10 +170,21 @@ export class PieChart extends React.Component<IPieChartProps, any> {
...
-        return d3.pie<DataType>().sort(null).value((d): any => d.value);
+        return d3.pie<IPieChartData>().sort(null).value((d): any => d.value);
     }
 
     private createColorPallette() {
         return d3.scaleOrdinal(d3.schemeCategory10);
     }
+
+    private _checkNameLen(name: string) : string {
+        if (name.length > 5) {
```
hbagarich: can we use text-overflow: ellipsis for these cases?

mpisacic: I'm not sure but probably, I'll try it

PR #11, opened 2017-02-21
Progress bar component

mpisacic:


src/components/ProgressBar/ProgressBar.props.ts
```diff
@@ -0,0 +1,14 @@
+export interface IProgressBarProps {
+    title: string;
+    width: number;
+    height: number;
+    percentage: string;
```
hbagarich: can this percentage be number as well?

vinkobedek: i would really like two values. one for the current value, one for max value. And then the progress bar to automatically calculate the percentage. if i already have a percentage then i would just set the max value to 100.

vinkobedek: and of course the default would be 100 if i do not supply a value for max

src/showcase.tsx
```diff
@@ -322,6 +323,7 @@ export class Index extends React.Component<any, any> {
                     width={330} 
                     height={200} 
                 ></LineChart>
+                <ProgressBar title={'RAM'} width={400} height={20} data={{total: 15999, current: 12560}} percentage={'79 %'}></ProgressBar>
```
vinkobedek: oh, i see, i can send total and current, but why do i have to specify the percentage then? What is the point of IUsage

mpisacic: IUsage is used so that internally component can calculate the size of progress and percentage is used for displaying it alongside progress. You're right both values seem a bit redundant, so I could either pass a total and current values and calculate percentage in class or just pass percentage.

src/components/ProgressBar/ProgressBar.props.ts
mpisacic: @hbagarich percentage is currently only used for display but we could have either only percentage or total and used values, but for me total and used seems like better option as it is more flexible.

PR #12, opened 2017-02-22
Dashboard component pull request

DanicaPivalica:
Still needs some improvements, but the basic components are here.


src/components/CompactDashboard/CompactDashboard.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+
+    public render() { 
+        let {title, farms} = this.props;
+        let classname = classNames({ [this.props.className]: this.props.className !== undefined });
+        return (
+            <div className={classname}>
+                {
+                    this.props.isVertical &&
+                    <div className="compact-dashboard-container vertical">
```
hbagarich: This can be seperate component?

src/components/CompactDashboard/CompactDashboard.tsx
```diff
@@ -0,0 +1,197 @@
...
+                                        width={width}
+                                        />
+                                )}
+                            </AutoSizer>
+                        }
+                    </div>
+                }
+                {
+                    !this.props.isVertical &&
+                    <div className="compact-dashboard-container ">
```
hbagarich: this can be separate component as well

src/components/CompactDashboard/CompactDashboard.tsx
```diff
@@ -0,0 +1,197 @@
...
+                                    <Collection
+                                        filt={this.props.filter}
+                                        ref={(reference) => {
+                                            this.collection = reference;
+                                        }}
+                                        cellCount={this.props.farms.length}
+                                        cellRenderer={this._renderRow}
+                                        cellSizeAndPositionGetter={function (index) {
+                                            return this.cellSizeAndPositionGetter(width, index);
+                                        }.bind(this)}
```
hbagarich: this binds the function every time it renders. you can bind this in constructor or using autobind decorator

src/components/CompactDashboard/CompactDashboard.tsx
```diff
@@ -0,0 +1,197 @@
...
+                                    <List
+                                        filt={this.props.filter}
+                                        height={height}
+                                         ref={(reference) => {
+                                            this.list = reference;
+                                        }}
+                                        rowCount={farms.length}
+                                        rowHeight={function (index) {
+                                            return this.calculateRowHeight(width, index);
+                                        }.bind(this)}
```
hbagarich: same problem here. we should bind this in contructor or use autobind decorator

src/components/CompactServer/CompactServer.tsx
```diff
@@ -0,0 +1,88 @@
...
+        return (
+            
+            <div className={ className }>
+               
+                    <span className={'server-title'} title={this.props.serverName} >{name}
+                        <span className={'server-close'} onClick={this.closeServer}>&times;</span>
+                    </span>
+                {
+                    this.props.roleList.length > 0 &&
+                    <hr/>
```
hbagarich: this hr can go before TagContainer component?

src/components/ServerDetails/ServerDetails.Props.ts
```diff
@@ -0,0 +1,51 @@
...
+    networkData: INetworkData;
+    diskData: IDiskData;
+    memoryData: IMemoryData;
+    headerData: IHeaderData;
+    cpuData: ICpuData;
+}
+
+export interface IHeaderData {
+    serverName: string;
+    fqdmServerName: string;
```
hbagarich: it should be fqdn

src/components/ServerDetails/ServerDetails.tsx
```diff
@@ -0,0 +1,119 @@
...
+                    
+                    <div className={'tl-disk-container'} style={{display: 'inline-block', margin: '0px 5px 0px 5px'}}>
+                        <Icon className="disk" iconName={'icon-LoadWithErrors'} style={{padding: '8px 5px 2px 5px'}}></Icon>
+                        <div className={'tooltip'}>
+                            {this._diskUsageList(this.props.headerData.diskData.disks)}
+                        </div>
+                    </div>
+                    {users}
+                </div>
+                <Details serverType={'Windows 2008 R2'} serverIcon={'icon-World'} serverRoles={
```
hbagarich: these serverTyoe and serverRoles props should be passed from the outside. 
Is this server icon class valid?

src/components/TagContainer/TagContainer.tsx
```diff
@@ -0,0 +1,81 @@
...
+                                <Icon iconName={tag.iconName}></Icon>
+                             }
+                             {
+                                 tag.display.length < 7 &&
+                                 <span>
+                                      {tag.display}
+                                </span>
+                            }
+                             {
+                                 tag.display.length > 7 &&
```
hbagarich: can we use text-overflow: ellipsis for these cases?

src/components/CompactServer/CompactServer.tsx
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    render() {
+         let isCritical = this.props.status === ServerStatus.Critical;
+         let isWarning = this.props.status === ServerStatus.Warning;
+         let isOK = this.props.status === ServerStatus.OK;
+         
+
+        let name = this.props.serverName;
+        if (name.length > 15) {
```
hbagarich: can we use text-overflow: ellipsis?

src/components/CompactDashboard/CompactDashboard.tsx
DanicaPivalica: I don't think it should. It woudn't serve any purpose other than 'cleaning-up' this look of the class. It won't be reusable due to the fact that each project will need to implement it's custom logic in handling react-virtualized components (if there is the need for some other components to be placed inside)

src/components/CompactDashboard/CompactDashboard.tsx
DanicaPivalica: Got it. Will change, didn't notice at first

src/components/CompactServer/CompactServer.tsx
DanicaPivalica: Yes :D There were some problems, but I've managed to get it right. Will be in the next commit.

src/components/CompactServer/CompactServer.tsx
DanicaPivalica: It must go before. I'm not sure what's the issue here

src/components/ServerDetails/ServerDetails.Props.ts
DanicaPivalica: This is Martin's component that I merged with my branch for further development. But He managed to change it, so I'm guessing it will be up soon

src/components/TagContainer/TagContainer.tsx
DanicaPivalica: In this case it's a bit more tricky because we don't have this design fully responsive. We have agreed on the fixed width of the component, but it's not the best solution... Also, this won't be checked that much as the previous case would be if we hadn't changed it. So I agree that it's not nearly the best option, but I would leave it for now.

src/components/CompactServer/CompactServer.tsx
hbagarich: we have two identical checks (this.props.roleList.length > 0). I was wondering if we can put all the code behind the single check?

src/components/TagContainer/TagContainer.tsx
hbagarich: my main concern is if the different font is used, this logic may produce unwanted behavior

PR #14, opened 2017-02-23
Server tile component

mpisacic:


src/components/ServerDetails/ServerDetails.Props.ts
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
+
+export interface IServerDetailsProps {
+    serverId: string;
+    serverStatusClass: string;
+    serverName: string;
+    fqdmServerName: string;
```
hbagarich: this should be serverFqdn

src/components/ServerDetails/ServerDetails.tsx
```diff
@@ -0,0 +1,81 @@
...
+            </div>
+        );
+    }
+
+    private _dismiss() {
+        this.props.onDismiss(this.props.serverId);
+    }
+
+    private _checkNameLen(name: string) : string {
+        if (name.length > 15) {
```
hbagarich: can we change this logic to use css property text-overflow: ellipsis. Different font and font sizes may cause unexpected behavior here

PR #20, opened 2017-03-02
Added on/off text to toggle switch.

mpisacic:


mpisacic:
Yes it can, but only problem is lack of space for text in current implementation.

hbagarich:
aha, ok. lets try it then without additional properties

hbagarich:
additionally, can this component now function in two modes? one with text, one without?

mpisacic:
Fixed, if props onText and offText are passed to component component will show it, if nothing is passed on then no text will be rendered.

PR #21, opened 2017-03-09
Bar chart component

mpisacic:


src/components/BarChart/BarChart.props.ts
```diff
@@ -0,0 +1,19 @@
+export interface IBarChartProps {
+    id: string;
+    barColor?: string;
+    hovColor?: string;
+    data: Array<IBarChartData>;
+    tipText?: (data: IBarChartData) => string;
+    xAxisFormat?: () => string;
+    width?: number;
+    height?: number;
```
hbagarich: why is there no minHeight/maxHeight property?

src/components/BarChart/BarChart.scss
```diff
@@ -0,0 +1,43 @@
...
+    display: inline-block;
+    border: 1px solid lightgrey;
+
+    -webkit-box-sizing: border-box; /* Safari/Chrome, other WebKit */
+    -moz-box-sizing: border-box;    /* Firefox, other Gecko */
+    box-sizing: border-box;  
+
+    svg { display: block; }
+
+    .domain {
```
hbagarich: should these classes be more specific?

src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,238 @@
...
+    private svg;
+    private container;
+
+    constructor(props: IBarChartProps) {
+        super(props);
+
+        this.x = this.generateX();
+        this.y = this.generateY();
+
+        // if (this.props.isResponsive) {
```
hbagarich: can we delete this comment?

src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,238 @@
...
+        //     this.width = window.innerWidth / 2;
+        //     this.height = 0.7 * this.width;
+        // }
+    }
+
+    public componentWillReceiveProps(nextProps: IBarChartProps) {
```
hbagarich: do we need this method?

src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    public shouldComponentUpdate(nextProps: IBarChartProps, nextState: null) {
+        if (this.props.id !== nextProps.id ||
+            this.props.width !== nextProps.width ||
+            this.props.height !== nextProps.height ||
+            this.props.barColor !== nextProps.barColor) {
+                return true;
+            }
+        if (!this.arraysEqual(this.props.data, nextProps.data)) { return true; }
```
hbagarich: this is too costly, especially if there is a lot of data in these arrays. can we use shallow compare here instead?

src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,238 @@
...
+            if ((data1.argument as Date).getTime() !== (data2.argument as Date).getTime()) { return false; }
+        }
+        if (data1.argument !== data2.argument) { return false; }
+        if (data1.frequency !== data2.frequency) { return false; }
+        return true;
+    }
+
+    public componentDidMount() {
+        this.init();
+        if (this.props.isResponsive) { window.addEventListener('resize', () => this._onResize()); }
```
hbagarich: we should remove this event lisnter on componentDidUnmount

src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,238 @@
...
+    public render() {
+        return <div className={'bar-chart-component'} ref={'container'}></div>;
+    }
+
+    private init() {
+        const xAxisClass = classNames('x-axis', this.props.id);
+        const yAxisClass = classNames('y-axis', this.props.id);
+
+        this.container = this.createContainer();
+        this.container.append('g').attr('class', xAxisClass).call(this.generateXAxis())
```
hbagarich: can you add some comments as to what is happening in this code. The d3 is powerful, but can be cryptic sometimes xD

src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,238 @@
...
+    }
+
+    private redraw() {
+        this.rescale(this.props.width, this.props.height);
+    }
+
+    private _onResize() : any {
+        let width = window.innerWidth / 2;
+        width = width < this.props.minWidth ? this.props.minWidth : width;
+        const height = width * 0.7;
```
hbagarich: maybe some comments here as well. Why these magic numbers, 0.7, 2?

src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,238 @@
...
+    private generateXAxis() {
+        return d3.axisBottom(this.x).tickPadding(10); // .tickFormat(d3.timeFormat(this.props.xAxisFormat()));
+    }
+
+    private generateYAxis() {
+        return d3.axisLeft(this.y).tickSizeInner(-this.width).tickPadding(5);
+    }
+
+    private generateBars(container: any) {
+        container.selectAll('.bar').data(this.props.data).enter()
```
hbagarich: can we get some comments here as well?

src/components/BarChart/BarChart.tsx
```diff
@@ -0,0 +1,238 @@
...
+        isResponsive: false,
+        onClick: () => { return; }
+    };
+
+    public refs: {
+        [argument: string]: (Element);
+        container: (HTMLInputElement);
+    };
+
+    private margin = { top: 50, right: 50, bottom: 50, left: 50 };
```
hbagarich: maybe some of these private properties can go into component state?

src/components/BarChart/BarChart.tsx
mpisacic: For some reason shallow compare doesn't work and re-rendering is always fired

src/components/BarChart/BarChart.tsx
mpisacic: This event listener makes component responsive but should be replaced.

src/components/BarChart/BarChart.tsx
mpisacic: These props just hold references to selections placing them in state will just means that there could be some unnecessary re-rendering

src/components/BarChart/BarChart.props.ts
mpisacic: Component scales based on width and aspect ratio so there is no need of min/max props for height

src/components/BarChart/BarChart.tsx
mpisacic: It would be a good idea if these magic numbers are replaced with constants

src/components/BarChart/BarChart.tsx
hbagarich: are you sure you are passing down the props that shallow compare know how to compare?
Deep comparing the arrays in onShouldComponentUpdate servers no purpose because it is very costly operation running frequently

src/components/BarChart/BarChart.tsx
hbagarich: I get it, but we should be able to safely remove it once the component is dismounted

PR #30, opened 2017-03-23
ServerTile popup on CompactDashboard

MatijaA:


src/components/CompactDashboard/CompactDashboard.tsx
```diff
@@ -31,19 +39,31 @@ function sortFarmServers(ob1: { status: number, name: string }, ob2: { status: n
     }
 }
 
-
 function checkFilter(filter: string, serverName: string): boolean {
     return serverName.toLowerCase().trim().indexOf(filter.toLowerCase().trim()) !== -1;
 }
 
-export class CompactDashboard extends React.Component<ICompactDashboardProps, any> {
+export class CompactDashboard extends CommonComponent<ICompactDashboardProps, any> {
```
hbagarich: we moved from inheriting commoncomponent to either React.Component or React.PureComponent if there is no need for custom should component update

PR #37, opened 2017-03-27
Spinner centered horizontally, font change.

mpisacic:


src/components/Spinner/Spinner.tsx
```diff
@@ -16,13 +16,13 @@ export class Spinner extends React.Component<ISpinnerProps, any> {
...
                    { 'spinner-large': type === SpinnerType.large })
-                }>
-                </div>
-                { label && (
-                    <div className={'spinner-label'}>{ label }</div>
-                )}
+                }/>
+                { label &&
+                    <div
+                        className={classNames('spinner-label', { 'normal': type === SpinnerType.normal }, { 'large': type === SpinnerType.large })}
```
hbagarich: instead of 'normal' can you use spinner-size-normal and the same approach for large?

PR #40, opened 2017-04-05
NavigationGroup component

iVergot:
Simple navigation element which consists of a header, followed by a vertical list of navigation element - link. Each element on click returns it's id (number)


src/components/NavigationGroup/NavigationGroup.scss
```diff
@@ -0,0 +1,58 @@
+@import '../../index.scss';
+
+.navigationGroupContainer{
```
hbagarich: Can you change the name of HTML classes so they are named according to this convention http://wiki.c2.com/?KebabCase

src/components/NavigationGroup/NavigationGroup.tsx
```diff
@@ -0,0 +1,38 @@
+import * as React from 'react';
+import { INavigationGroupProps } from './navigationGroup.Props';
+import { autobind } from '../../utilities/autobind';
+import './NavigationGroup.scss';
+import * as classNames from 'classnames';
+
+export class NavigationGroup extends React.Component<INavigationGroupProps, void> {
```
hbagarich: It would be better if navigation group extends React.PureComponent, as it will automatically compare passed props.

package.json
```diff
@@ -17,46 +17,46 @@
   "author": "",
```
hbagarich: These changes to package.json are not necessary, as we have npmshrnikwrap, but maybe it is clearer to change package.json as well :)

PR #42, opened 2017-04-06
Feat/breadcrumbs

mpisacic:
Breadcrumbs component redesigned with changed props and behaviour.


src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,185 +1,138 @@
...
+        const paths = this.state.currentPath.map((item, index) => {
+                const iconName = item.selected ? 'icon-arrow_down_right' : 'icon-arrow_right';
+                return (
+                    <li className={'breadcrumbs-list-item'} key={index} >
+                        <Dropdown
+                            ref={this.setDropdownReference}
+                            dropdownType={DropdownType.customDropdown} 
+                            icon={iconName} 
+                            onClosed={() => this.closeMenu(item)} 
+                            onOpenedClick={(willOpen) => this.openMenu(item, willOpen)}
```
hbagarich: Do the breadcrumb component needs to track this?

src/components/Dropdown/Dropdown.Props.ts
```diff
@@ -20,6 +20,7 @@ export interface IDropdownProps {
   calloutClassName ?: string;
   layerClassName ?: string;
   onClosed?: () => void;
+  onOpenedClick?: (opened?: boolean) => void;
```
hbagarich: Is this event needed?

src/components/Breadcrumbs/Breadcrumbs.tsx
mpisacic: Yes, for displaying menu.

src/components/Dropdown/Dropdown.Props.ts
mpisacic: It is needed by breadcrumbs component, maybe in some other use cases.

PR #45, opened 2017-04-10
DateTimePicker component

mpisacic:


package.json
```diff
@@ -22,10 +22,12 @@
     "classnames": "2.2.5",
     "css-element-queries": "0.4.0",
     "d3": "4.5.0",
+    "moment": "2.18.1",
```
hbagarich: You need to rebuild npm shrinkwrap file when adding new packages

PR #46, opened 2017-04-11
Fix/dashboard group

mpisacic:


src/components/LineChart/LineChart.tsx
```diff
@@ -32,21 +32,22 @@ export class LineChart extends React.PureComponent<ILineChartProps, any> {
...
+            isParentMounted: false,
+            chartId: 'line-' + Guid.raw()
         };
     }
 
     public render() {
-        const props = objectAssign({}, this.props, {width: this.state.width, height: this.state.height});
-        const componentClass = classNames('line-chart-component', this.props.id, this.props.className);
-        const titleClass = classNames('line-chart-title', this.props.id, this.props.className);
+        const props = objectAssign({}, this.props, { width: this.state.width, height: this.state.height, id: this.state.chartId });
```
hbagarich: You can remove the ID property out of the props if you are assigning the id from the state to props. This why you are basically changing the props in the render method

package.json
```diff
@@ -22,6 +22,7 @@
     "classnames": "2.2.5",
     "css-element-queries": "0.4.0",
     "d3": "4.5.0",
+    "guid": "0.0.12",
```
hbagarich: If you are adding new package to quick-react-ts, you have to rebuild npm shrinkwrap file

PR #47, opened 2017-04-12
Feat/breadcrumbs

mpisacic:


src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,154 +1,119 @@
...
-                            onClick={() => this.props.onPathClick(item.url)}
-                        >{item.name}</a>
-                    </li>
-                );
-            }
-        );
-
+    public render() {
+        const items = this.getDisplayItemsFromProps(this.props);
+        const rootItem = this.getFirstLevelFromProps(this.props);
```
hbagarich: Do you need root item here?

src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -1,154 +1,119 @@
...
+                    displayName: item.displayName,
+                    url: '/' + item.key,
+                }
+            );
+        }
+        let item : ICurrentPathItem = {
+            name: 'Home',
+            iconName: 'icon-home',
+            children: children,
+            url: '/'
```
hbagarich: It would be nice if IconName and Url properties are passed as props to Breadcrumbs component

src/components/Breadcrumbs/BreadcrumbsItem.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { Label } from '../Label/Label';
+import { Icon } from '../Icon/Icon';
+import { Dropdown } from '../Dropdown/Dropdown';
+import { autobind } from '../../utilities/autobind';
+import { DropdownType } from '../Dropdown/Dropdown.Props';
+import * as classNames from 'classnames';
+import { IBreadcrumbsItemProps, IBreadcrumbsChild } from './BreadcrumbsItem.Props';
+import './BreadcrumbsItem.scss';
+
+export class BreadcrumbsItem extends React.Component<IBreadcrumbsItemProps, any> {
```
hbagarich: Can you please change name of this class (and all connected) to BreadcrumbItem

src/components/Breadcrumbs/Breadcrumbs.tsx
mpisacic: That was left behind from older implementation, so it can be removed.

src/components/Breadcrumbs/BreadcrumbsItem.tsx
mpisacic: Do you propose some other name, maybe BreadcrumbsEntry or something similar?

src/components/Breadcrumbs/Breadcrumbs.tsx
hbagarich: Can you remove it?

src/components/Breadcrumbs/Breadcrumbs.Props.ts
```diff
@@ -1,11 +1,15 @@
 import * as React from 'react';
 import { Breadcrumbs } from './Breadcrumbs';
+import { IBreadcrumbChild } from './BreadcrumbItem.Props';
 
 export interface IBreadcrumbsProps extends React.Props<any> {
     items: IBreadcrumbItem[];
     className?: string;
     url: string;
-    onPathClick: (path: string) => void;  // return constructed url
+    homeName?: string;
```
hbagarich: Can this be named homeIconName, this is bit ambiguous.

PR #50, opened 2017-04-19
Fix/round values

mpisacic:
Small fixes to various components


package.json
```diff
@@ -22,6 +22,7 @@
     "css-element-queries": "0.4.0",
     "d3": "4.5.0",
     "guid": "0.0.12",
+    "lodash": "4.17.4",
```
hbagarich: you should rebuild npmshrink wrap file so npm can pick up this library as well

PR #51, opened 2017-04-19
Fix/dashboard display

mpisacic:
Removed empty spaces when searching in dashboard. Also added search by status with query.


src/components/DashboardHeader/DashboardHeader.tsx
```diff
@@ -8,14 +8,27 @@ import { Icon } from '../Icon/Icon';
 import * as classNames from 'classnames';
 import './DashboardHeader.scss';
 
+function debounce(func, wait, immediate?) {
```
vinkobedek: lodash already has a debounce function. https://lodash.com/docs/4.17.4#debounce. better to use that?

src/components/TileDashboard/TileDashboard.tsx
```diff
@@ -28,17 +29,26 @@ export class TileDashboard extends React.Component<ITileDashboardProps, any> {
...
-        if (this.props.filter !== prevProps.filter && this.list) {
+        if ((this.props.filter !== prevProps.filter && this.list) || prevProps.farms !== this.props.farms) {
             this.list.recomputeRowHeights();
         }
     }
 
-    render() {
-        let { farms } = this.props;
+    public componentWillReceiveProps(nextProps: ITileDashboardProps, nextState: any) {
+        const filteredFarms = filterFarms(nextProps.farms, nextProps.filter);
```
vinkobedek: maybe update the state only if the .farms prop is different?

src/components/TileGroup/TileGroup.tsx
```diff
@@ -6,15 +6,20 @@ import * as classNames from 'classnames';
...
 
 import './TileGroup.scss';
 
 export class TileGroup extends React.PureComponent<ITileGroupProps, void> {
 
     public render(): JSX.Element {
         const { farm } = this.props;
-        const servers = farm.servers.filter((server) => { return filterServerByName(this.props.filter, server.name); }).sort(sortServersByStatusAndName);
+        let servers;
+        if (this.props.filter.indexOf('status:') !== -1) {
```
vinkobedek: how's the performance with a lot of servers?

vinkobedek: also, why do this in render and not in componentwillreceiveprops only if the filter or farms differ

src/components/DashboardHeader/DashboardHeader.tsx
```diff
@@ -26,7 +22,7 @@ export class DashboardHeader extends React.Component<IDashboardHeaderProps, any>
                         <Icon className={'add-farm'} iconName={'icon-add'} onClick={this.props.onAddFarmClick} title={'Add'} />
                     }
                 </span>
-                <Search onSearch={this.props.onSearch} onChange={this.props.onChanged} value={this.props.filter} />
+                <Search onSearch={this.props.onSearch} onChange={_.debounce(this.props.onChanged, 250)} />
```
vinkobedek: i believe that the debounce should occur within the Search component, so if it is used elsewhere we do not have the same problem

PR #52, opened 2017-04-20
Added click event on window

marijanaSpajic:
Added click event on window and close left navigation when you click outside the component


src/components/LeftNavigation/LeftNavigation.tsx
```diff
@@ -4,18 +4,48 @@ import { ILeftNavigationProps, ILeftNavigationOption } from './LeftNavigation.Pr
...
+
     constructor(props) {
         super(props);
         this.state = { isOpen: false, selectedIndex: this.getSelectedIndex(this.props.options) };
     }
 
+    public componentDidMount() {
+        let target = this._leftNavElement;
+        this._setTargetWindowAndElement(target);
+        this._events.on(this._targetWindow, 'click', this._dismissOnClickOutsideComponent, true);
```
hbagarich: It would be nice to unsubscribe from this event in componentWillUnmount

PR #54, opened 2017-04-20
typescript upgrade to 2.2.2

MatijaA:


package.json
```diff
@@ -37,8 +37,11 @@
     "@types/node": "7.0.5",
     "@types/react": "15.0.1",
     "@types/react-dom": "0.14.20",
-    "@types/react-virtualized": "8.9.2",
-    "awesome-typescript-loader": "1.0.0",
+    "@types/reselect": "2.0.27",
```
hbagarich: You need to rebuild npmshrinkwrap file after updating the package.json file

MatijaA: Updated shrinkwrap and yarn.lock

PR #61, opened 2017-05-04
Server grid dashboard component

MatijaA:


package.json
```diff
@@ -24,27 +24,29 @@
...
+    "lodash.groupby": "4.6.0",
     "node-sass": "3.13.0",
     "object-assign": "4.1.1",
     "react": "15.4.2",
     "react-datetime": "^2.8.9",
     "react-dom": "15.4.2",
+    "react-draggable": "^2.2.3",
     "react-redux": "5.0.2",
     "react-virtualized": "9.7.3",
     "redux": "3.6.0",
```
mpisacic: Can you remove react-redux and redux dependencies because they are not used in project?

MatijaA: good idea, I'll remove them

package.json
```diff
@@ -25,14 +25,14 @@
     "guid": "0.0.12",
     "lodash": "4.17.4",
     "node-sass": "3.13.0",
-    "object-assign": "4.1.1",
     "react": "15.4.2",
     "react-datetime": "^2.8.9",
     "react-dom": "15.4.2",
-    "react-redux": "5.0.2",
+    "react-draggable": "^2.2.3",
```
hbagarich: Can you just remove ^ here, we have better control of versions that way

PR #63, opened 2017-05-05
Compact server without role is same height as with it.

mpisacic:


src/components/CompactServer/CompactServer.tsx
```diff
@@ -29,13 +34,18 @@ export class CompactServer extends React.PureComponent<ICompactServerProps, any>
...
-                        <hr/>
-                        <TagContainer title={''} tags={this.props.roles}/>
-                    </div>
-                }
+                <div>
+                    <hr/>
+                    {
+                        this.props.roles.length > 0 ? 
+                        (
+                            <TagContainer title={''} tags={this.props.roles} />
```
vinkobedek: wouldn't it be easier to give the tagcontainer, or more specifically the **edit-tags** class a minimum height?

PR #70, opened 2017-05-15
Fix for grouping & min column width

MatijaA:
The grid didn't work properly when no grouping column or minWidth was selected.
Removed autobind.


src/components/ServerGridDashboard/ServerGridDashboard.tsx
```diff
@@ -118,9 +118,7 @@ export class ServerGridDashboard extends React.Component<IServerGridDashboardPro
         this.state = {
             rows: this.transformFarmToRows(props.farms, props.filter),
             expandedRows: {},
-            groupBy: ['FarmName'],
-            sortColumn: 'ServerName',
```
hbagarich: Why did you remove sort configuration?

MatijaA: The sort configuration is in the Grids state and has no use in ServerGridDashboard

PR #71, opened 2017-05-15
Offline server

hbagarich:
Added support for offline servers to DetailedGroup and ServerDashboardGrid


src/components/DetailedServerTile/DetailedServerTile.scss
```diff
@@ -9,8 +9,9 @@ $counters-title-font-size: 12px;
 $counters-title-font-color: $basic-text-color;
 $server-status-color: $basic-text-color;
 
-.server-details {
+.detailed-server-details {
```
vinkobedek: we need to go deeper :D

PR #73, opened 2017-06-01
Feat/quick grid drag and drop

MatijaA:
changed on row hover method 
added Drag & Drop toolbar for grouping
groupBy column reordering inside the toolbar 
sort changed to support simultaneous groupBy & column sorting
removed autobind and arrow functions inside props

### DIFF
--- a/src/components/QuickGrid/DataSelectors.ts
+++ b/src/components/QuickGrid/DataSelectors.ts
@@ -13,29 +13,19 @@ const getGroupBySortDirection = (state: IQuickGridState, props: IQuickGridProps)
 const getColumns = (state: IQuickGridState, props: IQuickGridProps) => props.columns;
 
 const sortRows = (rows: Array<any>, sortColumn: string, sortDirection: string, groupedColumn: Array<string>, columns: Array<GridColumn>, groupBySortColumn: string, groupBySortDirection: string) => {
-    let column = columns.filter(x => x.valueMember === sortColumn)[0];
     let sortedChain = _(rows).chain();
-
-    if (sortDirection && sortColumn) {
-        if (column && column.sortByValueGetter) {
-            sortedChain = sortedChain.sortBy((row) => column.sortByValueGetter(row, sortDirection));
-        } else {
-            sortedChain = sortedChain.sortBy((row) => row[sortColumn]);
-        }
-        if (sortDirection === 'DESC') {
-            sortedChain = sortedChain.reverse();
-        }
-    }
-
     if (groupedColumn.length > 0) {
-        if (groupBySortColumn !== undefined && groupBySortColumn !== '') {
-            sortedChain.sortBy((row) => row[groupBySortColumn]);
+        const groupSortDir: string = groupBySortDirection === 'DESC' ? 'desc' : 'asc';
+        const groupSortBy = (groupBySortColumn !== undefined && groupBySortColumn !== '') ? groupBySortColumn : groupedColumn[0];
+        if (sortColumn) {
+            const columnSortDir: string = sortDirection === 'DESC' ? 'desc' : 'asc';
+            sortedChain = sortedChain.orderBy<any>([groupSortBy, sortColumn], [groupSortDir, columnSortDir]);
         } else {
-            sortedChain.sortBy((row) => row[groupedColumn[0]]);
+            sortedChain = sortedChain.orderBy(groupSortBy, groupSortDir);
         }
-        if (groupBySortDirection === 'DESC') { // TODO: test
-            sortedChain = sortedChain.reverse();
-        }
+    } else if (sortColumn) {
+        const sortDir = sortDirection === 'DESC' ? 'desc' : 'asc';
+        sortedChain = sortedChain.orderBy(sortColumn, sortDir);
     }
     return sortedChain.value();
 };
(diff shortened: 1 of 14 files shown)


src/components/QuickGrid/DataSelectors.ts
```diff
@@ -8,34 +8,31 @@ const getGroupBy = (state: IQuickGridState, props: IQuickGridProps) => props.gro
...
 
-const sortRows = (rows: Array<any>, sortColumn: string, sortDirection: string, groupedColumn: Array<string>, columns: Array<GridColumn>) => {
-    let column = columns.filter(x => x.valueMember === sortColumn)[0];
-    let sortedChain = _(rows).chain();
-
-    if (sortDirection && sortColumn) {
-        if (column && column.sortByValueGetter) {
-            sortedChain = sortedChain.sortBy((row) => column.sortByValueGetter(row, sortDirection));
+const sortRows = (rows: Array<any>, sortColumn: string, sortDirection: string, groupedColumn: Array<string>, columns: Array<GridColumn>, groupBySortColumn: string, groupBySortDirection: string) => {
+    const columnSortDir: string = sortDirection === 'DESC' ? 'desc' : 'asc';
```
hbagarich: you can just call toLowerCase here? or even change grid API so it uses enum instead of string? The same goes for the lower call.

src/components/QuickGrid/GroupByToolbar.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+class GroupByToolbarInner extends React.PureComponent<IGroupByProps, IGroupByState> {
+    constructor(props: IGroupByProps) {
+        super(props);
+        this.state = {
+            groupBy: props.groupBy
+        };
+    }
+    componentWillReceiveProps(nextProps) {
+        this.setState({ ...this.state, groupBy: nextProps.groupBy });
```
hbagarich: this will update the state even if the previous props are the same?

src/components/QuickGrid/QuickGrid.Props.ts
```diff
@@ -6,23 +6,31 @@ export interface IQuickGridProps {
...
-    highlightHoverRow?: boolean;
     onSelectedRowChanged?: (selectedRowIndex: number) => void;
     onRowDoubleClicked?: (row: any) => void;
+    displayGroupContainer?: boolean;
+
+    sortColumn?: string;
+    sortDirection?: 'ASC' | 'DESC';
+    onGroupByChanged?: (groupBy: Array<string>) => void;
+
+    groupBySortColumn?: string;
```
hbagarich: we are still supporting only one level of grouping?

src/components/QuickGrid/QuickGrid.Props.ts
```diff
@@ -6,23 +6,31 @@ export interface IQuickGridProps {
...
+    groupBySortDirection?: 'ASC' | 'DESC';
```
hbagarich: if the underlying library (lodash?) wants 'asc' and 'desc' should we also change to that? But it seems to me that usage of enum would be much cleaner.

src/components/QuickGrid/QuickGridHeader.tsx
```diff
@@ -1,84 +1,96 @@
...
+        this.columnMinWidths = this.getColumnMinWidths(props.headerColumns);
+    }
+
+    getColumnMinWidths(columns) {
+        return columns.map((col) => { return col.minWidth || 0; });
     }
+
     componentWillReceiveProps(nextProps) {
         this.setState({ ...this.state, columnWidths: nextProps.columnWidths });
+        this.columnMinWidths = this.getColumnMinWidths(nextProps.headerColumns);
```
hbagarich: As per documentation componentWillRecieveProps can be called event if the new props are not passed. We should check if the props changed before calculating widths?
Also we will change state event if the props are the same?

src/components/QuickGrid/QuickGrid.Props.ts
MatijaA: The grid now supports multiple levels of grouping, you can sort by one group and column

src/components/QuickGrid/DataSelectors.ts
MatijaA: I have added SordDirection enum, it's definitely better than using string definitions in all interfaces

src/components/QuickGrid/QuickGridHeader.tsx
MatijaA: Great suggestion, I committed a fix to check for changes on componentWillReceiveProps

src/components/QuickGrid/GroupByToolbar.tsx
```diff
@@ -0,0 +1,153 @@
...
+class GroupByToolbarInner extends React.PureComponent<IGroupByProps, IGroupByState> {
+    constructor(props: IGroupByProps) {
+        super(props);
+        this.state = {
+            groupBy: props.groupBy
+        };
+    }
+
+    componentWillReceiveProps(nextProps) {
+        if (!_.isEqual(nextProps.groupBy, this.props.groupBy)) {
```
hbagarich: this performs deep comparison? although it should not be problem in this specific case, i think we are much safer with shallow compare in react life cycle events

src/components/QuickGrid/QuickGridHeader.tsx
```diff
@@ -1,89 +1,103 @@
...
+        this.columnMinWidths = this.getColumnMinWidths(props.headerColumns);
+    }
+
+    getColumnMinWidths(columns) {
+        return columns.map((col) => { return col.minWidth || 0; });
     }
+
     componentWillReceiveProps(nextProps) {
-        this.setState({ ...this.state, columnWidths: nextProps.columnWidths });
+        if (!_.isEqual(nextProps.columnWidths, this.props.columnWidths)) {
```
hbagarich: Lodash isEqual performs deep comparison, but it seems to me that we could use shallow comparison. It is faster and more inline with the React

PR #76, opened 2017-06-20
tree filter implementation

MatijaA:
resizable tree filter with single or multi select, search, select-all, select all filtered (only with a flat list)

### DIFF
--- a/src/components/TreeFilter/TreeFilter.scss
+++ b/src/components/TreeFilter/TreeFilter.scss
@@ -1,0 +1,106 @@
+@import '../../index.scss';
+
+ul {
+    margin: 0;
+    list-style: none;
+}
+
+.item {
+    height: 20px;
+    display: inline-block;
+    flex-direction: column;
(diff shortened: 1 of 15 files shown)


src/components/TreeFilter/TreeFilter.scss
```diff
@@ -0,0 +1,106 @@
...
+    font-family: $font-family-base;
+    font-size: $font-size-m;
+}
+
+// checkbox
+$Checkbox-label-size: 20px;
+$Checkbox-transition-duration: 200ms;
+$Checkbox-transition-timing: cubic-bezier(.4, 0, .23, 1);
+
+.checkbox {
```
vinkobedek: @MatijaA didn't this override the default checkbox style?

MatijaA: the style was copied from the checkbox component. I have changed the class name to avoid conflict

PR #77, opened 2017-06-21
Feat/time picker keyboard input

ivansicic:
Add keyboard input to TimePicker 
 -useKeyboardInput prop specifies whether keyboard input is enabled or not, default value is false
 -time is changed on every keyUp
 -arrows increase and decrease time

### DIFF
--- a/src/components/DateTimePicker/DateTimePicker.tsx
+++ b/src/components/DateTimePicker/DateTimePicker.tsx
@@ -13,7 +13,7 @@ export interface IDateTimePickerProps {
     is24HourFormat: boolean;
     includeTime?: boolean;
     isValidDate?: (currentDate: any, selectedDate: any) => boolean;
-    keyboardTimeInput?: boolean;
+    useKeyboardForTimeInput?: boolean;
     onTimeSelectionChanged: (selectedDateTime: Date) => void;
 }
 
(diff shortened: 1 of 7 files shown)


src/components/DateTimePicker/DateTimePicker.tsx
```diff
@@ -13,6 +13,7 @@ export interface IDateTimePickerProps {
     is24HourFormat: boolean;
     includeTime?: boolean;
     isValidDate?: (currentDate: any, selectedDate: any) => boolean;
+    keyboardTimeInput?: boolean;
```
vinkobedek: useKeyboardInput IMO sounds better. maybe even useKeyboardForTimeInput.

src/components/DateTimePicker/TimePicker.tsx
```diff
@@ -52,22 +58,93 @@ export class TimePicker extends React.Component<ITimePickerProps, void> {
...
+            if (hour > 12) {
+                hour = 12;
+            }
+            if (isPeriodPm) {
+                if (hour < 12) {
+                    hour += 12;
+                }
+            } else {
+                if (hour === 12) {
+                    hour = 0;
```
vinkobedek: this is a bit weird. you can never have 0 AM. it's 12 AM

ivansicic: If a displayed time is 12 AM, it's 0 in 24-hour format and needs to be set as 0 in outside Date representation (props.hour).

vinkobedek: i think that the problem is that i am not quite sure if this function return the display value [1-12] or [0-24], the actual value. The **if hour > 12** line leads me to believe that we are working within the [1-12] range as an output, but for the AM part of the time it sets 0 which can never be. So i'm a bit confused. the this.props.OnTimeChangeCall should work with [0-23] since we do not know if it is AM or PM, but from the code, it will be clipped to 12 hours.

ivansicic: The **if hour > 12** line clips the value because input went over the allowed max value, so it's being handled here, setting it to max value for 12-hour format. After that, conversion to 24-hour format follows. If the period is PM, 12 hours should be added to the input value (except when it is 12 PM), and if AM, 12 AM will be converted to 0.  
Function always converts to [0-24], but starts with correcting input errors.

vinkobedek: ok, reading the code now, it all makes sense. don't know what i thought before.

Issue #74, opened 2017-06-07
Some QuickGrid props could be optional


vinkobedek:
- groupBy, no need to specify an empty array
- rowHeight, we could just take 20 as a default
- headerHeight, we could also provide some default

PR #83, opened 2017-07-05
QuickGrid sort by every groupBy column, bugfixes

MatijaA:
- Added an example page for QuickGrid 
- Fixed grid resizing
- Every grouped column is sorted bey default
- added optional props - as per issue #74 
- Fixed text overflow on column headers
- Remove group by columns by dragging them out of the toolbar


src/components/QuickGrid/QuickGrid.Props.ts
```diff
@@ -4,34 +4,34 @@ export enum SortDirection {
     Ascending,
     Descending
 }
+
+export interface IGroupBy {
+    column: string;
+    sortDirection: SortDirection;
```
hbagarich: can we add capability to define custom sort here?

src/components/QuickGrid/DataSelectors.ts
```diff
@@ -24,22 +22,27 @@ const sortRows = (rows: Array<any>, sortColumn: string, sortDirection: SortDirec
     }
 
     if (groupedColumn.length > 0) {
-        const groupSortDir: string = groupBySortDirection === SortDirection.Descending ? 'desc' : 'asc';
-        const groupSortBy = (groupBySortColumn !== undefined && groupBySortColumn !== '') ? groupBySortColumn : groupedColumn[0];
+        let sortColumns = [];
+        let sortDirections = [];
+        for (let groupColumn of groupedColumn) {
```
hbagarich: can we add capability to define custom sort rules for grouped columns?

MatijaA: Added the use of custom sort function on grouped columns, if the function is defined

PR #89, opened 2017-07-07
Feat/tree filter selected count

MatijaA:
- added footer with a number of selected items/ total items
- added a reset to default filter settings button 
- the callout is downsized if there are a few elements 
- added a partially selected icon
- changed to a different resizing library - callout resizable from all directions 
- fixed bug - partial select on filtered values
- select all working on filtered values


src/components/Icon/Icon.tsx
```diff
@@ -6,10 +6,9 @@ import { getNativeAttributes, htmlElementAttributes } from '../../utilities/attr
 import './Icon.scss';
 
 export const Icon: (props: IIconProps) => JSX.Element = (props: IIconProps) => {
-
-    let size = props.size;
-    let customIcon = props.iconName === '';
-    let iconClass = props.iconName;
+    // let size = props.size;
+    const customIcon = props.iconName === '';
+    // let iconClass = props.iconName;
```
hbagarich: You can delete these commented lines?
It does seem that these changes will not affect current usage of Icon component.

PR #99, opened 2017-07-31
Optional dropdown open/close controlled from props

mpisacic:
There was no way (as I am familiar) to close custom dropdown, so this prop can be used when building custom dropdowns with controlled open / close state.


src/components/Dropdown/Dropdown.tsx
```diff
@@ -215,12 +226,18 @@ export class Dropdown extends React.Component<IDropdownProps, any> {
         this.setState({
           isOpen: !this.state.isOpen
         });
+        if (this.props.onMenuToggle) {
+          this.props.onMenuToggle(false);  // return next state of opened
```
MatijaA: shouldn't it be: this.props.onMenuToggle(!this.state.isOpen)

PR #109, opened 2017-08-22
Added fullName prop to disk info and sorted disk by name

marijanaSpajic:


src/components/ServerTile/DisksInformation/DisksInformation.tsx
```diff
@@ -66,7 +67,7 @@ export class DisksInformation extends React.PureComponent<IDisksInformationProps
                             <div key={index} className={classNames(
                                 { 'status-warning': data.status === ServerStatus.Warning },
                                 { 'status-critical': data.status === ServerStatus.Critical }
-                            )}>{data.name}: {data.used}/{data.capacity} {data.usageUnit}</div>
+                            )}>{data.fullName}: {data.used}/{data.capacity} {data.usageUnit}</div>
```
vinkobedek: from the interface fullName is optional, there should be additional logic that tests if fullname is provided.

PR #111, opened 2017-08-23
Feat/wizard

vedranserences:
Wizard component


package.json
```diff
@@ -1,82 +1,84 @@
...
+    "object-assign": "^4.1.1",
+    "react": "15.4.2",
+    "react-datetime": "2.8.9",
+    "react-dnd": "2.4.0",
+    "react-dnd-html5-backend": "2.4.1",
+    "react-dom": "15.4.2",
+    "react-draggable": "2.2.3",
+    "react-numeric-input": "2.0.7",
+    "react-resizable-box": "2.0.4",
+    "react-router": "3",
```
hbagarich: we should not depend on react router in quick-react-ts

src/components/Wizard/Wizard.tsx
```diff
@@ -0,0 +1,132 @@
...
+      });
+      this.props.onPageRender(this.state.currentStep - 1);
+      this.props.onPageLeave(this.state.currentStep, this.state.currentStep - 1);
+      this.props.onPageEnter(this.state.currentStep - 1, this.state.currentStep);
+    }
+  }
+
+  private _cancelCreateScript(e) {
+    e.preventDefault();
+    hashHistory.push('/');
```
hbagarich: it would be better to use onCancel callback instead of directly using hashHistory

PR #115, opened 2017-09-19
GroupBy now properly works with valueMember and dataMember

sglogovic:
ValueMember in quickGrid is used for sort/group value while dataMember for display. GroupBy is now properly shows dataMember while grouping

### DIFF
--- a/src/components/QuickGrid/DataSelectors.ts
+++ b/src/components/QuickGrid/DataSelectors.ts
@@ -47,6 +47,6 @@ const getSortedRows = createSelector(getInputRows, getSortColumn, getSortDirecti
         return sortRows(rows, sortColumn, sortDirection, groupBy, columns);
     });
 
-export const getRowsSelector = createSelector(getSortedRows, getGroupBy, getExpandedRows, (rows, groupedColumns, expandedRows = {}) => {
-    return groupRows(rows, groupedColumns, expandedRows);
+export const getRowsSelector = createSelector(getSortedRows, getGroupBy, getExpandedRows, getColumns, (rows, groupedColumns, expandedRows, columns = {}) => {
+    return groupRows(rows, groupedColumns, expandedRows, columns);
 });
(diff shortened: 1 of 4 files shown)


src/components/QuickGrid/DataSelectors.ts
```diff
@@ -47,6 +47,6 @@ const getSortedRows = createSelector(getInputRows, getSortColumn, getSortDirecti
         return sortRows(rows, sortColumn, sortDirection, groupBy, columns);
     });
 
-export const getRowsSelector = createSelector(getSortedRows, getGroupBy, getExpandedRows, (rows, groupedColumns, expandedRows = {}) => {
-    return groupRows(rows, groupedColumns, expandedRows);
+export const getRowsSelector = createSelector(getSortedRows, getGroupBy, getExpandedRows, getColumns, (rows, groupedColumns, expandedRows, columns = {}) => {
```
MatijaA: is columns = {} intentionally ?

sglogovic: No, it should stay with expandedRows

src/components/QuickGrid/rowGrouper.ts
```diff
@@ -1,11 +1,13 @@
 import * as _ from 'lodash';
-import { GroupRow, IGroupBy } from './QuickGrid.Props';
+import { GroupRow, IGroupBy, GridColumn } from './QuickGrid.Props';
 class RowGrouper {
     groupByColumns: Array<IGroupBy>;
+    columns: Array<GridColumn>;
     expandedRows: any;
-        constructor(groupByColumns, expandedRows) {
+        constructor(groupByColumns, expandedRows, columns) {
         this.groupByColumns = groupByColumns.slice(0);
```
hbagarich: We raised the TypeScript version, so we can use spread operator instead of slice, [...groupByColumns]

sglogovic: Replaced

PR #120, opened 2017-09-27
fixed dashboard tile height calculation

MatijaA:
fixed bug: when server measure (cpu or disk) was 0 the tile would display "--"
fixed server tile & compact tile height calculation depending on server roles

### DIFF
--- a/src/components/ServerGridDashboard/ServerGridDashboard.tsx
+++ b/src/components/ServerGridDashboard/ServerGridDashboard.tsx
@@ -50,13 +50,13 @@ const gridColumns: Array<GridColumn> = [{
             case ServerStatus.Warning:
                 modifier = sortDirection === SortDirection.Ascending ? '1' : '3';
                 break;
-                case ServerStatus.OK:
+            case ServerStatus.OK:
                 modifier = sortDirection === SortDirection.Ascending ? '2' : '2';
                 break;
             case ServerStatus.Offline:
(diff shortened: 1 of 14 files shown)


src/components/ServerGridDashboard/ServerGridDashboard.tsx
```diff
@@ -69,7 +69,7 @@ const gridColumns: Array<GridColumn> = [{
     dataMember: 'CPUData',
     width: 100,
     minWidth: GRID_CELL_MIN_WIDTH,
-    cellFormatter: (cellData) => { return <div className={GetClassForStatus('', cellData.status) + ' server-dashboard-grid-cell-content'} > {cellData.usage ? cellData.usage + '%' : '--'}</div>; },
+    cellFormatter: (cellData) => { return <div className={GetClassForStatus('', cellData.status) + ' server-dashboard-grid-cell-content'} > {cellData.usage != null ? cellData.usage + '%' : '--'}</div>; },
```
hbagarich: What if the cellData.usage is undefined or empty?

MatijaA: If cellData.usage is undefined the cellData.usage != null will return true, that case is correct.
usage is a number converted to string, it shouldn't be possible that it's empty. 
Is this check necessary?

hbagarich: You have to format to '--' if there is no usage for the specific cell, so I think the check is required.

PR #126, opened 2017-10-11
Feat/quick grid expand and collapse all rows

ivansicic:
Add expand and collapse all to QuickGrid

 - add expand all and collapse all buttons
 - rename expandedRows to collapsedRows and change it to string (group key) array
 - add expandAll and collapseAll functions in QuickGrid.tsx


src/components/QuickGrid/QuickGrid.scss
```diff
@@ -1,352 +1,362 @@
...
+    margin-bottom: 15px;
+    float: left;
+    border: 1px solid #c8c8c8;
+    .group-drop-empty {
+        margin: 7px auto;
+        width: 200px;
+        display: block;
+        color: #4b4949;
+    }
+    > .expand-collapse-buttons {
```
ivansicic: This is the only real change in the file. Each line had a single space padding, will be removed now.

src/components/QuickGrid/rowGrouper.ts
```diff
@@ -5,19 +5,19 @@ import { groupBy } from '../../utilities/array';
...
         this.columns = [...columns];
     }
 
-    isRowExpanded(columnName, name) {
+    isRowExpanded(name) {
         let isExpanded = true;
-        let expandedRowGroup = this.expandedRows[columnName];
-        if (expandedRowGroup && expandedRowGroup[name]) {
-            isExpanded = expandedRowGroup[name].isExpanded;
+        let index: number = this.collapsedRows.indexOf(name, 0);
```
vinkobedek: will this be a performance issue?

ivansicic: Doesn't seem to be so at a data sample of 100k rows. Object (Map) would be faster, but working with an array is much cleaner, in terms of code readability. 
If objects were to be used, assignment would look like this: collapsedRows[groupKey] = {}, or collapsedRows[groupKey] = true,  but the value would never be used. Also removing properties from this structure (having 'true' or an empty object as value) wouldn't make sense to someone reading this code later, but keeping all the groupKeys as collapsed rows and assigning them boolean values doesn't either. Corrections and suggestions are welcome! :D

PR #127, opened 2017-10-12
QuickGrid optimization

MatijaA:
- raised react-virtualized version
- rows with lower case members are in row selector, not grid state
- added private field finalGridRows, this removes the row selector call on every cellRenderer
-

### DIFF
--- a/src/components/QuickGrid/QuickGrid.tsx
+++ b/src/components/QuickGrid/QuickGrid.tsx
@@ -238,7 +238,7 @@ export class QuickGridInner extends React.Component<IQuickGridProps, IQuickGridS
     }
 
     onActionItemClick = (option, rowIndex) => {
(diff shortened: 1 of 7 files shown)


src/components/QuickGrid/QuickGrid.tsx
```diff
@@ -273,19 +237,20 @@ export class QuickGridInner extends React.Component<IQuickGridProps, IQuickGridS
         );
     }
 
+    onActionItemClick = (option, rowIndex) => {
+        const { onActionSelected, actionIconName, actionItems } = this.props.gridActions;
```
ivansicic: Nice to pull the function out 👍 , but do you still need **actionIconName**?

MatijaA: You are correct, it's not needed. It was a case of quick C/P :)

PR #130, opened 2017-10-17
Add tooltips to QuickGrid

ivansicic:
- Grid cells, Header, Action cells and Footer now have tooltips enabled by default
 - cell tooltip shows cell data, action cell tooltip shows actions tooltip
 - tooltips can be disabled via tooltipsEnabled prop in QuickGrid
 - action cells (column) tooltip can be passed via actionsTooltip prop in QuickGrid
 - cells that have their own formatter function have to take of the tooltip on their own


src/components/QuickGrid/HeaderColumn.tsx
```diff
@@ -43,7 +48,7 @@ class HeaderColumnInner extends React.PureComponent<IHeaderColumnProps, void> {
                     <div className="header-text">
                         <span
                             key="label"
-                            title={text}
+                            {...tooltipsEnabled ? {title: text} : {} }
```
vinkobedek: won't this create a new object each time?

ivansicic: You're right. Ty 👍

PR #141, opened 2017-11-02
Added title to treeview item

josipGrgic11235:
Added a "title" property to TreeViewItemProps; a string which, on hover, will be displayed as a tooltip above the specific TreeViewItem.

### DIFF
--- a/src/components/Treeview/TreeviewItem.Props.ts
+++ b/src/components/Treeview/TreeviewItem.Props.ts
@@ -24,4 +24,5 @@ export interface ITreeviewItem {
     parentId?: string;
     hoverOverBtn?: Array<IHoverOverBtn>;
     className?: string;
+    title?: string;
 }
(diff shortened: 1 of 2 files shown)


mpisacic:
ITreeViewItem has text property which is then displayed in TreeView, would it be better if title were that text itself, or we gain something better from this approach ?

josipGrgic11235:
It's quite likely that sometimes we wouldn't like the title to be displayed which isn't doable without an additional property. If that were the case, in my opinion there are only two options: 
1) introduce a boolean flag, something like "showTitle" and, depending on it's value, do/don't show the text property.

2) have a "titleText" string property, and show it if it has a value, don't if it's unset (undefined).

If we choose the 2nd approach, we are more flexible with the decision what text to show in the tooltip. Noted, in more than some occasions, the title and text props will be the same.

We could also introduce showTitle as well as titleText props. If that were the case we could do something like:
<span ... title = {showTitle ? (titleText ? titleText : text) : ''}
That gives us the ability only to set the showTitle flag to true if we wish to set the tooltip text to be the same as the text property

PR #144, opened 2017-11-03
TextField error tooltip

marijanaSpajic:
- added new error tooltip in TextField  component
- added prop hasTitleBorder in TreeFilter component

### DIFF
--- a/src/components/TextField/TextField.tsx
+++ b/src/components/TextField/TextField.tsx
@@ -302,7 +302,7 @@ export class TextField extends React.Component<ITextFieldProps, ITextFieldState>
                 className="tooltip-error"
                 showTooltip={this.state.isFocused}
                 directionalHint={DirectionalHint.bottomLeftEdge}>
-                <div>
+                <div className="text-field-textarea-error-content">
                     <textarea
                         { ...textAreaProps }
                         id={this._id}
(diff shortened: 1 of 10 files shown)


src/components/TextField/TextField.tsx
```diff
@@ -120,10 +123,9 @@ export class TextField extends React.Component<ITextFieldProps, ITextFieldState>
                 {label && <Label htmlFor={this._id}>{label}</Label>}
                 {iconClass && <i className={iconClass}></i>}
                 {multiline ? this._renderTextArea() : this._renderInput()}
-                {(description || errorMessage) &&
+                {description &&
```
hbagarich: can the description also be implemented as tooltip to save space around the component?

src/components/TextField/TextField.tsx
```diff
@@ -278,37 +280,89 @@ export class TextField extends React.Component<ITextFieldProps, ITextFieldState>
...
-                className={this._fieldClassName}
-                onFocus={this._onFocus}
-                onBlur={this._onBlur}
-            />
+            <Tooltip
+                content={this.props.errorMessage}
+                className="tooltip-error"
+                showTooltip={this.state.isFocused}
+                directionalHint={DirectionalHint.bottomLeftEdge}>
+                <div>
```
hbagarich: I feel like this div should have assignable className so it can be styled if there is need?

PR #146, opened 2017-11-06
Feat/message box component

mpisacic:
MessageBox component

### DIFF
--- a/src/components/MessageBox/MessageBox.tsx
+++ b/src/components/MessageBox/MessageBox.tsx
@@ -85,7 +85,12 @@ export class MessageBox extends React.Component<IMessageBoxProps, {}> {
                 }
                 {
                     errorMessage !== undefined && errorMessage !== '' &&
-                    <Tooltip containerClass="operation-error-tooltip-container" content={errorMessage} directionalHint={DirectionalHint.rightCenter}>
+                    <Tooltip
+                        className="tooltip-error"
+                        containerClass="operation-error-tooltip-container"
+                        content={errorMessage}
+                        directionalHint={DirectionalHint.rightCenter}
+                    >
                         <Icon className="operation-error" iconName="icon-usklicnik" />
                     </Tooltip>
                 }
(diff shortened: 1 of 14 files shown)


src/components/MessageBox/MessageBox.tsx
```diff
@@ -0,0 +1,100 @@
...
+                    <Icon iconName={iconName} />
+                    {message}
+                </div>
+                {
+                    isLoading && <Spinner className="message-box-spinner" />
+                }
+                {
+                    errorMessage !== undefined && errorMessage !== '' &&
+                    <Tooltip containerClass="operation-error-tooltip-container" content={errorMessage} directionalHint={DirectionalHint.rightCenter}>
+                        <Icon className="operation-error" iconName="icon-usklicnik" />
```
hbagarich: can this iconName be either a prop or something more appropriate?

src/components/MessageBox/MessageBox.tsx
```diff
@@ -0,0 +1,100 @@
...
+                <div className="message-box-container">
+                    <Icon iconName={iconName} />
+                    {message}
+                </div>
+                {
+                    isLoading && <Spinner className="message-box-spinner" />
+                }
+                {
+                    errorMessage !== undefined && errorMessage !== '' &&
+                    <Tooltip containerClass="operation-error-tooltip-container" content={errorMessage} directionalHint={DirectionalHint.rightCenter}>
```
hbagarich: can you please break this line into multiple ones?

iVergot:
@mpisacic what if I want to display message box without status icon (messageLevel enum)?

hbagarich:
@iVergot is this something that we want/anticipate to happen? message boxes have icons?

mpisacic:
@iVergot if you dont provide level as prop you wont get icon

iVergot:
@mpisacic you will, default value for level prop is MessageLevel.Information, you have to pass null to remove the status Icon.
@hbagarich I don't know, that's why i asked 😄

mpisacic:
@iVergot some idea is that message box have that three states, so if you want no icon hack it ;)

hbagarich:
I think we should go with message boxes with icons based on MessageLevel. I think that is expected of the MessageBox component. If the end user really wants no icon, there is hackish solution, but by default the MessageBox should show icons.

iVergot:
since there is close button defined in component properties, should message box close itself on close button click after executing user defined method?

mpisacic:
@iVergot  close button has associated onClose hook (or callback) which defines behavior on close from outside of component, which in general makes this component a controlled one, in some use case I could define that when I press close button some form gets cleared (ie. dispath some action) and then close box with setting its. isOpen prop to false, also we can't just close box on completion because box doesn't know when its associated operation is completed (it is coming from "outside")

PR #149, opened 2017-11-08
Added new dropdown type, refactored code 

iVergot:
Added new dropdown type,  refactored code to reduce unnecessary reredering


iVergot:
@hbagarich no, @marijanaSpajic already fixed it :)

PR #155, opened 2017-11-10
Added onDateSelectionChanged as optional on CustomDateRange component

marijanaSpajic:


examples/src/CustomDateRange.tsx
```diff
@@ -54,15 +56,31 @@ export class Index extends React.Component<any, any> {
...
+            invalidErrorMessage: ''
+        });
+    }
+
+    @autobind
+    private onDateSelectionChanged(selectedStartDate: Date, selectedEndDate: Date) {
+        const invalidSelectedDate = moment(selectedStartDate) <= moment(selectedEndDate).subtract(3, 'day');
+        this.setState({
+            invalidDateRangeSelected: invalidSelectedDate,
+            invalidErrorMessage: invalidSelectedDate ? 'The time period between Start and End date must not be more than 3 days.' : ''
```
hbagarich: This seems a bit specific? Can we take the same approach as in TextField component, to use invalidErrorMessage prop to define invalid date selection?

PR #156, opened 2017-11-16
Implemented Compare component

sglogovic:
Added compare component to quick-react-ts.

Component displays left and right grid with shared scroll based on row and column data.
Component props :

- sourceRows - Collection of row data for left grid
- targetRows - Collection of row data for right grid
- columns - Collection of column data for both grids
- differences - Collection of rowIndexes, columnIndexes and differenceTypes 

NOTE : 
This component does not compare data form source and target row, differences property recieves what is different and what is not.

image


src/components/Compare/CompareComponent.tsx
```diff
@@ -0,0 +1,380 @@
...
+                                    <div className={'grid-component-container'} style={{height : height - 50}}>
+                                        <div className={'compare-single-grid-wrapper'}>
+                                            <GridHeader
+                                                allColumns={this.columns}
+                                                headerColumns={this.columns}
+                                                columnWidths={this.state.columnWidths}
+                                                width={this.gridComponentWidth(width)}
+                                                scrollLeft={scrollLeft}
+                                                className={'grid-component-header'}
+                                                groupBy={[]}
```
hbagarich: this cause rerender always. Maybe ommit this prop or use const?

sglogovic: Made groupBy optional in QuickGridHeader as suggested.

sglogovic:
Made groupBy optional in QuickGridHeader as suggested.

sglogovic:
Added defaultProperty value for groupBy to avoid rerendering.

PR #158, opened 2017-11-22
Bug/bugfixes

marijanaSpajic:
- Dashboard height fixes
- Dialog height fixes
- Left Navigation position fixes
- export DirectionalHint
- click on the server when grouping is none on Dashboard component
- added icon prop to Label component
- removed some css


src/components/Dashboard/Dashboard.scss
```diff
@@ -1,17 +1,23 @@
...
-.viewport-height {
-    height: 90vh;
+.dashboard {
     width: 100%;
+    height: 100%;
 
-    &:focus {
-        outline: none;
+    .viewport-height {
+        height: calc(100% - 84px);
```
hbagarich: this magic number is always used?

src/components/Layers/Layer.scss
```diff
@@ -11,6 +11,7 @@
     visibility: visible;
     position: absolute;
     width: 100%;
+    z-index: 9999;
```
hbagarich: this seems dangerous, especially for dropdowns and callouts inside layers?

src/components/Dashboard/Dashboard.scss
marijanaSpajic: Yes, it is because dashboard header has defined height of 84px

src/components/Layers/Layer.scss
marijanaSpajic: I agree, but the z-index has to be here so I will change him to 10

src/components/Dashboard/Dashboard.scss
hbagarich: Ok :)

src/components/Layers/Layer.scss
hbagarich: Ok :)

PR #159, opened 2017-11-22
Feat/condition control

iVergot:
Added compare conditions control. Control is stateless and all events must be handled from outside container.
Control supports multiple level grouping of conditions. Each condition consists of logical operator (and - or), property name, compare condition and compare value.
Compare conditions are user defined, and compare value editor is generated according to property type.


src/components/ConditionSelector/Condition.tsx
```diff
@@ -0,0 +1,158 @@
...
+            selectedCompareConditionKey: option.key
+        };
+        if (this.props.compareConditionChanged) {
+           this.props.compareConditionChanged(state);
+        }
+    }
+    
+    renderItemEditor = (propertyType: PropertyTypeEnum,  additionalData: any) => {
+        switch (propertyType) {
+            case PropertyTypeEnum.String: {
```
hbagarich: It would be better if these cases would be either functions that return JSX elements or even better separate components (this seems a perfect use case for React Functional Components https://reactjs.org/docs/components-and-props.html )

src/components/ConditionSelector/Condition.tsx
```diff
@@ -0,0 +1,158 @@
...
+function collectSource(connect, monitor) {
+    return {
+        connectDragSource: connect.dragSource(),
+        connectDragPreview: connect.dragPreview(),
+        isDragging: monitor.isDragging()
+    };
+}
+
+@DragSource('condition', conditionSource, collectSource)
+export class Condition extends React.PureComponent <ConditionDefinition, any> {
```
hbagarich: It would be better to use an specific interface for state (or simply if {} if this component does not have state) instead of any. 
This way we leverage TypeScript features and nobody can mistakenly call this.setState.

src/components/ConditionSelector/ConditionGroup.Props.ts
```diff
@@ -0,0 +1,21 @@
...
+    Or,
+    Empty = 1000
+}
+
+export class ExpressionDefinitionTree {
+    id: string;
+    logicalOperator: LogicalOperatorTypeEnum;
+    conditionDefinition?: ConditionDefinition;
+    subExpressions?: ExpressionDefinitionTree[];    
+    isOver?: any;
```
hbagarich: is this a boolean?

src/components/ConditionSelector/ConditionSelector.tsx
```diff
@@ -0,0 +1,143 @@
...
+}
+
+export class ConditionSelectorContainerPorps {
+    conditionSelectorProps: ConditionSelectorPorps;
+    onDragDrop?: (newTree: ExpressionDefinitionTree) =>  void;
+}
+
+
+@DragDropContext(HTML5Backend)
+export class ConditionSelectorContainer extends React.PureComponent <ConditionSelectorContainerPorps, any> {
```
hbagarich: Can you extract this class to another file so we are consistent.

src/components/ConditionSelector/ConditionSelector.tsx
```diff
@@ -0,0 +1,143 @@
...
+                            />
+                        }
+                    </div>
+                </div>
+            </div>
+        );
+    }
+}
+
+export class ConditionSelectorContainerPorps {
```
hbagarich: It' better to use interfaces to define react props. Also there is a typo here, ConditionSelectorContainer**Porps** -> ConditionSelectorContainer**Props**

src/components/ConditionSelector/ConditionSelector.Props.ts
```diff
@@ -0,0 +1,20 @@
+import { ITreeviewItem } from '../Treeview';
+import { IExpressionDefinitionTree, LogicalOperatorTypeEnum, IConditionRowState } from './';
+
+export interface IConditionSelectorPorps {
```
hbagarich: Typo here, it is IConditionalSelector**Props*

src/components/ConditionSelector/Condition.Props.ts
```diff
@@ -0,0 +1,32 @@
...
+    classname?: string;
+    compareConditions?: IDropdownOption[];
+    hasIndent?: boolean; // default false 
+    indentSize?: number; // default 30
+    compareConditionChanged?: (conditionRowState: IConditionRowState) => void;
+    conditionValueChanged?: (conditionId: string, conditionValue: any) => void;
+    additionalData?: any;
+    conditionDragged?: (dragSource: any, dragTarget: any) => void;
+    connectDragSource?: any;
+    isDragging?: any;
```
hbagarich: is this boolean?

PR #181, opened 2017-12-22
Feat/hoverable btn added to virtualized tree view

kacavendadino:
- added HoverableButton HOC component that adds hoverable buttons to the given component
- added hoverable buttons to tree items in Virtualized Tree View

### DIFF
--- a/src/components/HoverableButton/HoverableButton.tsx
+++ b/src/components/HoverableButton/HoverableButton.tsx
@@ -1,0 +1,66 @@
+import * as React from 'react';
+import { TreeviewItemHoverBtn } from '../Treeview/treeviewItemHoverBtn';
+import { IHoverOverBtn } from '../Treeview';
+import './HoverableButtons.scss';
+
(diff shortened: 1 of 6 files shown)


src/components/HoverableButton/HoverableButton.tsx
```diff
@@ -0,0 +1,66 @@
...
+                hover: false
+            };
+        }
+    
+        render(): JSX.Element {
+            if (!hoverableProps.hoverOverBtn) {
+                return <Comp {...this.props} />;
+            }
+
+            return <span className="hoverable-items__container" onMouseEnter={this._onItemHover.bind(this)} onMouseLeave={this._onItemLeaveHover.bind(this)}>
```
vinkobedek: can we use autobind?

kacavendadino: @vinkobedek unfortunately no, when i tried it said that decorators can't be used in this context

PR #185, opened 2018-01-08
Feat/tree grid compact

hrestakd:
UPDATE:
New TreeGrid component which uses a modified version of QuickGrid to enable tree functionality inside a grid with the following features:
- Expandable tree structure of virtualized grid
- Working column sort on all levels
- Resizeable columns (with the exception of the collapse action column)
- Resizeable container with automatically adjusting column widths
- Added a "fixedWidth" property to the GridColumn interface

Existing QuickGrid component has been expanded with optional props for custom cell renderer, custom sorting and static columns.

TreeGrid usage is almost exactly the same as QuickGrid. Only Tree and Columns props are required.
(See examples for further info)


src/components/TreeGrid/TreeGrid.Props.ts
```diff
@@ -0,0 +1,28 @@
...
+    collapsedTreeNodes: Array<TreeNode>;
+    selectedRowIndex?: number;
+    columnsToDisplay: Array<GridColumn>;
+    sortColumn?: string;
+    sortDirection?: SortDirection;
+}
+
+
+export interface TreeNode { // extend this interface on a data structure to be used for row data
+    TreeId: string;
```
vinkobedek: Is this the Id of the node or the id of the whole tree? I presume just the node. In that case it would be better to have id as the name.

src/components/TreeGrid/TreeGrid.Props.ts
```diff
@@ -0,0 +1,28 @@
...
+    ParentId: string; // treeId of the parent node
+    IsExpanded: boolean;
+    leaves: Array<TreeNode>;
```
vinkobedek: leaves gives me a picture of only nodes on the lowest level. will the root node in a tree with 5 levels also have leaves.length > 0? If so then a better name would be children

src/components/TreeGrid/TreeGrid.Props.ts
```diff
@@ -0,0 +1,28 @@
...
+    selectedRowIndex?: number;
+    columnsToDisplay: Array<GridColumn>;
+    sortColumn?: string;
+    sortDirection?: SortDirection;
+}
+
+
+export interface TreeNode { // extend this interface on a data structure to be used for row data
+    TreeId: string;
+    ParentId: string; // treeId of the parent node
```
vinkobedek: why pascalcase?

src/components/TreeGrid/TreeGrid.Props.ts
hbagarich: Also we typically use camelCase convention for property naming. Can you please change all the properties to camelCase?

src/components/TreeGrid/TreeGrid.tsx
```diff
@@ -0,0 +1,254 @@
...
+        let icon = null;
+        
+        if (rowData.leaves.length <= 0 && rowData.IsExpanded) {
+            icon = null;
+            actionsTooltip = null;
+        } else {
+            icon = <Icon iconName={iconName} className="expand-collapse-action-icon" />;
+        }
+        const rowClass = 'grid-row-' + rowIndex;
+        const onMouseEnter = () => { this.onMouseEnterCell(rowClass); };
```
hbagarich: These arrow function may cause unnecessary rerenders. 
You could avoid them by using special component for the **div** you are returning and binding these functions.

src/components/TreeGrid/TreeGrid.tsx
```diff
@@ -0,0 +1,254 @@
...
+        const cellData = rowData[dataKey];
+        const rowClass = 'grid-row-' + rowIndex;
+        const className = classNames(
+            'grid-component-cell',
+            rowClass,
+            column.cellClassName,
+            { 'border-column-cell': notLastIndex },
+            { 'is-selected': rowIndex === this.state.selectedRowIndex });
+
+        const onMouseEnter = () => { this.onMouseEnterCell(rowClass); };
```
hbagarich: These arrow function may cause unnecessary rerenders.
You could avoid them by using special component for the div you are returning and binding these functions.

src/components/TreeGrid/TreeGrid.Props.ts
hrestakd: Changed

src/components/TreeGrid/TreeGrid.Props.ts
hrestakd: Indeed, changed

src/components/TreeGrid/TreeGrid.Props.ts
hrestakd: No idea :) changed

src/components/TreeGrid/TreeGrid.tsx
hrestakd: @hbagarich I made a new child component "CellElement". Can you just check it over real quickly to see if its properly implemented? 

Also when rendering the column element I am passing the result of ColumnElement function. Unsure if that is considered in the same context as passing the reference to the arrow function?

hrestakd: Also I'm unsure if the @autobind is needed in that child component but I added it anyway. If it's redundant I can remove it

src/components/TreeGrid/CellElement.tsx
```diff
@@ -0,0 +1,50 @@
...
+    @autobind
+    private _onMouseEnter() { 
+        if (this.props.onMouseEnter) {
+            this.props.onMouseEnter(this.props.rowClass);
+        }
+    }
+    @autobind
+    private _onMouseLeave() { 
+        if (this.props.onMouseLeave) {
+        this.props.onMouseLeave(this.props.rowClass);
```
hbagarich: Indentation here. :)

PR #186, opened 2018-01-09
Feat/left navigation hide menu

marijahlobik:
The leftNavigation component has:
- option to hide the menu burger 
- icons that are added to the prop otherOptions are positioned at the bottom


src/components/LeftNavigation/LeftNavigation.Props.ts
```diff
@@ -7,6 +7,7 @@ export interface ILeftNavigationProps extends React.Props<any> {
     className?: string;
     onClick?: (ev?: React.MouseEvent<HTMLElement>, item?: ILeftNavigationOption) => void;
     otherOptions?: ILeftNavigationOption[];
+    hideMenu?: boolean;
```
vinkobedek: hideMenu is more of a name for a function because it signals an action, something to do. Maybe use menuHidden? And is it really hiding the menu? or is it disabling the expand functionality?

marijahlobik: It is disabling the expand functionality for burger menu. menuHidden sounds right

PR #187, opened 2018-01-10
Refactor/async tree load

vinkobedek:
most of the implementation regarding the async loading feature was actually in the example and would have to be redone each time the feature was to be used. 

expandOrCollapseAsyncTreeItem is the new function that handles the tree reconstruction when items are requested.
```javascript
export function expandOrCollapseAsyncTreeItem(
getRoots: () => TreeItem[],
 treeItemToExpand: TreeItem, 
lookupTableGetter: () => TreeLookups,
 onTreeUpdated: (newRoots: TreeItem[]) => void,
 asyncItemsGetter: () => Promise<TreeItem[]>): void {
```
In the default case it behaves as a normal expand. But if the node has _hasChildren_ set to true and no children, then this signal a async load.

the tree roots are needed to reconstruct the tree as we want it to be immutable. Since we will be reconstructing it two times(start of the async action, end of the async action) we need to be able to get the latest values, hence the getRoots callback instead of the actual values.

lookupTableGetter, same deal as with tree roots.

Instead of returning the rebuilt tree we need to update it twice(start, end of action), so the onTreeUpdated callback was added.

asyncItemsGetter is the logic that actually retrieves the new data.

The reason for this convoluted logic is so that everything regarding the async load is limited to this component. Otherwise in every implementation we would have to watch out that we would need to trach which nodes are currently loading, which are loaded, etc.

Not entirely happy with this solution presented but i cannot for the life of me come up with anything better.

### DIFF
--- a/src/components/TreeFilter/VirtualizedTreeView.tsx
+++ b/src/components/TreeFilter/VirtualizedTreeView.tsx
@@ -182,7 +182,12 @@ export class VirtualizedTreeView extends React.PureComponent<IVirtualizedTreeVie
         }
         const onExpandClick = (event) => {
             event.stopPropagation();
-            this.props.onItemExpand(treeItem, this.props.lookupTableGetter);
+            if (this.props.onItemExpand) {
+                this.props.onItemExpand(treeItem, this.props.lookupTableGetter);
+            } else {
+                treeItem.expanded = !treeItem.expanded;
+                this._list.recomputeRowHeights();
+            }
         };
 
         const filterSelection = this.props.filterSelection;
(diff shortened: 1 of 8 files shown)


src/components/TreeFilter/VirtualizedTreeView.tsx
```diff
@@ -282,6 +292,27 @@ export class VirtualizedTreeView extends React.PureComponent<IVirtualizedTreeVie
...
+            height: this.props.rowHeight,
+            marginLeft: this.props.isSingleSelect ? 6 : 18
+        };
+            return (
+                <div
+                    key={loadingTreeNodeKey + '_Loading'}
+                    className="item-container loading-container"
+                    style={style}
+                >
+                    <Spinner className="tree-view-async-loading-spinner"
```
marijanaSpajic: The Spinner component has his own prop label so you can use it to write Loading...

PR #188, opened 2018-01-10
Feat/svg icons

rivisic:
- Added new svg icon system.
- Current Icon component changed.
- Usage example: ` <Icon iconName="svg-icon-*" width="16px" height="16px"/>` where * is icon name


src/components/Icon/Icon.scss
```diff
@@ -1,23 +1,26 @@
...
-	display: inline-block;
+    -webkit-font-smoothing: antialiased;
+    display: inline-block;
     font-family: icomoon;
     font-style: normal;
-	margin-right: 5px;
+    margin-right: 5px;
+}
+
+.svg {
```
vinkobedek: maybe nest the svg class inside the icon class to limit the scope to the icon component.

src/components/Icon/Icon.tsx
```diff
@@ -12,6 +12,14 @@ export const Icon: (props: IIconProps) => JSX.Element = (props: IIconProps) => {
             [props.iconName]: !customIcon
         }, [props.className]);
 
-    return <i { ...getNativeAttributes(props, htmlElementAttributes) } className={iconClassName} />;
+        if (props.iconName.startsWith('svg')) {
+            return <svg className = {'svg ' + props.iconName} width = {props.width} height= {props.height}>
```
vinkobedek: props.className is ignored for svg. Is this the desired behaviour?

src/components/Icon/Icon.scss
rivisic: Ok, but currently in the .icon class is many font-related properties which are not necessary for svgs.

src/components/Icon/Icon.tsx
rivisic: It will be changed

src/components/Icon/Icon.scss
vinkobedek: indeed, you are correct, i mistakenly thought that there was some parent element, but it always returned <i/> directoly. how about removing the svg class and the using the following selectors
i.icon, svg.icon.

in the icon component we would the return either return 
```<i { ...getNativeAttributes(props, htmlElementAttributes) } className={iconClassName} />;```
or
```<svg className={iconClassName} />;```


with this the iconClassName will work as before and the component will respect the props.className.

PR #206, opened 2018-01-25
Feat/button loader

marijahlobik:
ButtonLoader component is button with with Spinner, Succes and Error feature. 

http://take.ms/rURU9

In ButtonLoader.Props are defined properties - you can set flags:
 - isLoading - if true, you will se button with spinner
 - isSucces - if true, you will see green button with checkmark
 - isError - if true, you will see red button with error icon


mpisacic:
Basically what we did is wrapped different sorts of buttons based on some state (ie. loading, success) and created some sort of factory, which means every time state changes we'll spill out different component. Should be maybe better to incorporate this kind of behaviour inside button component, which will based on this props change display state, and not the whole component, and in general component wont be unmounted and remounted which could potentially be faster. How big of procedure that will be, what do you think?

PR #217, opened 2018-02-01
Added iconStyle prop to messageBox

rivisic:
iconStyle prop added to messageBox


src/components/MessageBox/MessageBox.Props.ts
```diff
@@ -15,6 +15,7 @@ export interface IMessageBoxProps {
     title?: string;
     level?: MessageLevel;
     iconName?: string;
+    iconStyle?: any;
```
vinkobedek: There is a type for this somewhere in the React typings. just f12 on the style attribute in some component to find out.

PR #218, opened 2018-02-05
Bug/check list and ppl picker fixes

kacavendadino:
- fix for text overflow for checkbox list items
- people picker suggestion box goes over elements below the control
- limited picker suggestion box height
- added help button to wizard

### DIFF
--- a/src/components/PeoplePicker/PeoplePicker.tsx
+++ b/src/components/PeoplePicker/PeoplePicker.tsx
@@ -23,6 +23,9 @@ export interface IPeoplePickerState {
 export class PeoplePicker extends React.PureComponent<IPeoplePickerProps, IPeoplePickerState> {
     private _field;
     private _delayedSearch;
+    public static defaultProps: Partial<IPeoplePickerProps> = {
+        noResultText: 'No Result'
+    };
 
     constructor(props) {
         super(props);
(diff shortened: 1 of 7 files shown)


src/components/PeoplePicker/PeoplePicker.tsx
```diff
@@ -65,20 +68,25 @@ export class PeoplePicker extends React.PureComponent<IPeoplePickerProps, IPeopl
...
 
+                    allSelected = allSelected && alreadySelected;    
                     return !alreadySelected && <Principal
                         key={principal.identifier}
                         principal={principal}
                         isSelected={false}
                         onSelect={this._onSuggestionClick}
                     />;
                 })}
+                {!this.props.loadingSuggestionList && (this.props.suggestionList.length === 0 || allSelected) && <div className="no-result">
```
vinkobedek: why are we showing no results text when allSelected? for that matter what is the purpose of allSelected, it starts as set to true and thats it, or am i missing something

kacavendadino: @vinkobedek this is just a check if in the suggestion list are only principals that are all already selected, because we are not showing principals that are already selected so this situation is the same as if a user typed a term that returns 0 principals

PR #230, opened 2018-02-16
tree data can now be any datasource

vinkobedek:
We started to use the TreeGrid component also for non-hierarchical data, as such it became weird to extend the treenode interface when you just have a simple list of items.

Now we can just pass any old array of objects and it should work.

### DIFF
--- a/src/models/TreeData.ts
+++ b/src/models/TreeData.ts
@@ -1,6 +1,6 @@
 export interface TreeNode { // extend this interface on a data structure to be used for row data    
     isExpanded?: boolean;
-    children: Array<TreeNode>;
+    children?: Array<TreeNode>;
     hasChildren?: boolean;
     iconName?: string;
     iconTooltipContent?: string;
@@ -47,7 +47,7 @@ export class TreeDataSource {
      * 
      * @param root warning: will be mutated and returned as ITreeDataSource
      */
-    constructor(input: TreeNode | TreeDataSource | Array<TreeNode>) {
+    constructor(input: TreeNode | TreeDataSource | Array<any>) {
         if (this.isDataSource(input)) {
             this.nodesById = input.nodesById;
             this.idCounter = input.idCounter;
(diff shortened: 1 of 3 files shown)


src/models/TreeData.ts
```diff
@@ -83,11 +83,11 @@ export class TreeDataSource {
...
     }
 
-    private isDataSource(input: TreeNode | TreeDataSource | Array<TreeNode>): input is TreeDataSource {
+    private isDataSource(input: TreeNode | TreeDataSource | Array<any>): input is TreeDataSource {
         return (<TreeDataSource>input).updateNode !== undefined;
     }
 
-    private isRootNodesArray(input: TreeNode | TreeDataSource | Array<TreeNode>): input is Array<TreeNode> {
+    private isRootNodesArray(input: TreeNode | TreeDataSource| Array<any>): input is Array<any> {
         return (<Array<TreeNode>>input).slice !== undefined;
```
kacavendadino: @vinkobedek 
Doesn't this cast negate the Array<any> above? If you send Array of type that isnt TreeNode it will return that root node is not array

src/models/TreeData.ts
```diff
@@ -47,7 +47,7 @@ export class TreeDataSource {
      * 
      * @param root warning: will be mutated and returned as ITreeDataSource
      */
-    constructor(input: TreeNode | TreeDataSource | Array<TreeNode>) {
+    constructor(input: TreeNode | TreeDataSource | Array<any>) {
         if (this.isDataSource(input)) {
             this.nodesById = input.nodesById;
```
kacavendadino: @vinkobedek 
Will this be handled if any doesnt have this fields? (will the TreeGrid know how to work if these fields are undefined)

src/models/TreeData.ts
vinkobedek: in the end, this is all just javascript, so this cast will succeed in runtime. this is why we have the need for these utility functions in the first place that determine of which object type it actually is. But it is definitely misleading, and i shall change it

src/models/TreeData.ts
vinkobedek: as disscused,  changed to to any to avoid confusion.

PR #241, opened 2018-02-27
Bug/people picker

zlatkoHrastic:
- People picker was redesigned
- Input box doesn't change size when principal is selected
- Fixed a bug that caused error tooltip to show up in the top left corner of the page
- Icons are now optional and can be added to people picker via functions:
    mapPrincipalToIcon?: (IPrincipal) => string;
    mapPrincipalToIconClass?: (IPrincipal) => string;

### DIFF
--- a/examples/MockData/peoplePickerData.ts
+++ b/examples/MockData/peoplePickerData.ts
@@ -1,4 +1,4 @@
-import { IPrincipal } from '../../src/components/PeoplePicker/Principal.Props';
+import { IPrincipal, PrincipalType } from '../../src/components/PeoplePicker/Principal.Props';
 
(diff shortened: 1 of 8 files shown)


examples/MockData/peoplePickerData.ts
```diff
@@ -7,3 +7,32 @@ export const peoplePickerData: IPrincipal[] = [
     { identifier: '4', displayName: 'FirstName4 LastName4', email: 'username4@domain4', type: 4 },
     { identifier: '5', displayName: 'FirstName5 LastName5', email: 'username5@domain5', type: 1 }
 ];
+export function MapIcon(principal: IPrincipal): string {
```
vinkobedek: camelCase?

PR #243, opened 2018-02-27
added feature to hide icons in dropdown

tomislav-m:


src/components/QuickGrid/QuickGridRowContextActionsHandler.tsx
```diff
@@ -11,6 +11,7 @@ import { DirectionalHint } from '../../index';
 export interface IQuickGridRowAContextActionsHandlerProps {
     onGetRowActions?: (rowIndex: number) => Array<ActionItem>;
     onActionClicked?: (rowIndex: number, action: ActionItem) => void;
+    hideDropdownActionIcons?: boolean;
```
vinkobedek: i think that you can avoid this property. I mean, here we have a dropdown component that accepts options that may or may not have icons. Why would i just not give the component options without the icons defined?

src/components/QuickGrid/QuickGrid.tsx
```diff
@@ -601,7 +601,7 @@ export class QuickGridInner extends React.Component<IQuickGridProps, IQuickGridS
         return (
             <div className={mainClass}>
                 <div className="hoverActions">
-                    <QuickGridRowContextActionsHandler ref={this._setRowContextActionsHandler} onGetRowActions={this.getRowContextActions} onActionClicked={this.onRowContextActionClicked} />
+                    <QuickGridRowContextActionsHandler ref={this._setRowContextActionsHandler} onGetRowActions={this.getRowContextActions} onActionClicked={this.onRowContextActionClicked} hideDropdownActionIcons={this.props.gridActions.hideDropdownActionIcons}/>
```
vinkobedek: in line with the comment about avoiding the dropdown hide icons property, this is also not needed

src/components/QuickGrid/QuickGridRowContextActionsHandler.tsx
```diff
@@ -119,52 +120,52 @@ export class QuickGridRowContextActionsHandler extends React.PureComponent<IQuic
...
+        if (renderDropDown) {
+            elements.push(mapAction(actions[0]));
+            elements.push(mapAction(actions[1]));
+            let actionOptions = [];
+            for (let i = 2; i < actions.length; i++) {
+                let tooltipInfo = actions[i].tooltip;
+                if (tooltipInfo && !tooltipInfo.directionalHint) {
+                    tooltipInfo = { ...tooltipInfo, directionalHint: DirectionalHint.leftCenter };
+                }
+                actionOptions.push({ key: actions[i].commandName, icon: actions[i].iconName, text: actions[i].name, tooltipInfo });
```
vinkobedek: this is the actual place where you can remove the icons from the hover menu, way simpler, simply do not set the icon prop :).

src/components/QuickGrid/QuickGrid.tsx
vinkobedek: on the second pass, i am wrong, this is actually needed, i mistakenly thought that the whole gridactions was passed down to the actionshandler. Live & learn.

src/components/QuickGrid/QuickGridRowContextActionsHandler.tsx
tomislav-m: you are right, done :D

src/components/QuickGrid/QuickGridRowContextActionsHandler.tsx
```diff
@@ -119,52 +120,51 @@ export class QuickGridRowContextActionsHandler extends React.PureComponent<IQuic
...
+        if (renderDropDown) {
+            elements.push(mapAction(actions[0]));
+            elements.push(mapAction(actions[1]));
+            let actionOptions = [];
+            for (let i = 2; i < actions.length; i++) {
+                let tooltipInfo = actions[i].tooltip;
+                if (tooltipInfo && !tooltipInfo.directionalHint) {
+                    tooltipInfo = { ...tooltipInfo, directionalHint: DirectionalHint.leftCenter };
+                }
+                actionOptions.push({ key: actions[i].commandName, icon: this.props.hideDropdownActionIcons === false ? actions[i].iconName : undefined, text: actions[i].name, tooltipInfo });
```
vinkobedek: what happens if this.props.hideDropdownActionIcons  is not set? the name leads me to believe that by default they will be shown. But the logic here if i am not mistaken makes it so that the are visible only when it is explicitly set to false.

tomislav-m: true

PR #246, opened 2018-03-01
Adding prop isRowNotSelectable - TreeGrid, QuickGrid

marijahlobik:
I added property isRowNotSelectable(boolean ) to TreeGrid.Props and QuickGrid.Props.

If you want to show data in TreeGrid but you don't want to select them onClick, you can set prop isRowNotSelectable as true. Also if you don't want that row has color on hover you can set the same prop in QuickGrid and nothing is going to happen on hover.


marijahlobik:
Ok I'll do it, make sense.

PR #248, opened 2018-03-07
Feat/add dropdown error

zlatkoHrastic:
Added ability to add error and error tooltip to dropdown

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -90,12 +90,13 @@ export class Dropdown extends React.PureComponent<IDropdownProps, IDropdownState
         const dropdownIconClassName = hasTitleBorder ? 'iconArrowWithBorder' : 'iconArrow';
 
         const selectionTextObj = this.getSelectionText(dropdownType, selectedOption);
-
+        const isError = this.props.validated === undefined ? false : !this.props.validated;
         const dropdownTitleClassName = classNames({
             'dropdown-title-border': hasTitleBorder,
             'dropdown-title': !hasTitleBorder,
             'arrow': showArrowIcon,
-            'text': selectionTextObj.hasText
+            'text': selectionTextObj.hasText,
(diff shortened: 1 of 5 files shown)


src/components/Dropdown/Dropdown.tsx
```diff
@@ -90,12 +90,13 @@ export class Dropdown extends React.PureComponent<IDropdownProps, IDropdownState
         const dropdownIconClassName = hasTitleBorder ? 'iconArrowWithBorder' : 'iconArrow';
 
         const selectionTextObj = this.getSelectionText(dropdownType, selectedOption);
-
+        const isError = this.props.validated === undefined ? false : !this.props.validated;
```
vinkobedek: you can put validated: true in defaultProps to avoid this check. ie. it could be used somewhere else in the future so we do not have to repeat this logic.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -121,30 +122,54 @@ export class Dropdown extends React.PureComponent<IDropdownProps, IDropdownState
...
-                        }
-                    </span>
+                    {isError &&
+                        < Tooltip
+                            content={this.props.validationErrorMessage}
+                            delayMs={500}
+                            directionalHint={DirectionalHint.rightCenter}
+                            className={'tooltip-error'}>
+                            <span className={dropdownTitleClassName} title={this.props.label}>
+                                {icon && (
```
vinkobedek: if i am not mistaken this is the same code as below. since it is a single element you can easily put it in some variable and use it in both places.

zlatkoHrastic: Moved the code to variable

src/components/Dropdown/Dropdown.tsx
zlatkoHrastic: Added default prop and changed the prop name for better understanding

PR #250, opened 2018-03-09
refactof LeftNavigation

marijahlobik:
In LeftNavigation.Props I removed otherOptions from props. 

I added position property in ILeftNavigationOption[]. You can define if your option is supposed to be down with LeftiNavigationOptionPositionEnum.Down.

[Example](https://monosnap.com/file/de3EAHepuruS5ziBZ22i4socIThRUL)

This refactor is because otherOptions weren't selected.


src/components/LeftNavigation/LeftNavigation.Props.ts
```diff
@@ -24,4 +23,10 @@ export interface ILeftNavigationOption {
     icon?: string;
     selected?: boolean;
     disabled?: boolean;
+    position?: LeftiNavigationOptionPositionEnum;
```
vinkobedek: typeo

PR #253, opened 2018-03-13
fix for undefined selected option

marijahlobik:
I added check if none option is selected (or there is no property selected = true)

### DIFF
--- a/src/components/LeftNavigation/LeftNavigation.tsx
+++ b/src/components/LeftNavigation/LeftNavigation.tsx
@@ -76,12 +76,14 @@ export class LeftNavigation extends CommonComponent<ILeftNavigationProps, any> {
         this.props.onClick(ev, item);
     }
 
-    private getSelectedId(options: Array<ILeftNavigationOption>) {
+    private getSelectedId(options: Array<ILeftNavigationOption>) : string {
         const selectedOptions = options.filter(option => { return option.selected ? option.id : null; });
-        
+
         if (selectedOptions.length > 0) {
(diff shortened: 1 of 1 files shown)


src/components/LeftNavigation/LeftNavigation.tsx
```diff
@@ -78,7 +78,10 @@ export class LeftNavigation extends CommonComponent<ILeftNavigationProps, any> {
 
     private getSelectedId(options: Array<ILeftNavigationOption>) {
         const selectedOptions = options.filter(option => { return option.selected ? option.id : null; });
-        return selectedOptions[0].id;
+        
+        if (selectedOptions.length > 0) {
+            return selectedOptions[0].id;
+        }
```
hbagarich: Can you explicitly return some value from function?

src/components/LeftNavigation/LeftNavigation.tsx
```diff
@@ -78,7 +78,10 @@ export class LeftNavigation extends CommonComponent<ILeftNavigationProps, any> {
 
     private getSelectedId(options: Array<ILeftNavigationOption>) {
```
hbagarich: Can you declare which type should this function return?

marijahlobik: I did it now

src/components/LeftNavigation/LeftNavigation.tsx
marijahlobik: I returned empty string (because id is defined as string)

PR #262, opened 2018-03-16
TreeGrid scrolls to selected item on selection change

mpisacic:
Added scroll to support on TreeGrid and QuickGrid so that QuickGrid can be scrolled to certain position. On top of that scrolling to was implemented on TreeGrid so that when selected item on tree  is changed tree automatically scroll to some position so that selected item will be visible.


src/components/QuickGrid/QuickGrid.scss
```diff
@@ -170,7 +170,8 @@
             }
 
             &.is-selected {
-                background-color: $disabled-field;
+                background-color: $selected-item-background;
+                font-weight: bold;
```
hbagarich: Why are we changing style?

mpisacic: so that selected item is better visible

src/components/TreeGrid/TreeGrid.tsx
```diff
@@ -1,23 +1,23 @@
...
+import { ITreeGridProps, ITreeGridState } from './TreeGrid.Props';
+import { getTreeRowsSelector } from './treeGridDataSelectors';
 
 export class TreeGrid extends React.PureComponent<ITreeGridProps, ITreeGridState> {
     public static defaultProps = {
         isNodeSelectable: true
     };
 
-    private _quickGrid: any;
+    private readonly _rowOverselect = 0;
```
vinkobedek: what does this do if it is always zero?

mpisacic: nothing much, can be removed

PR #265, opened 2018-03-20
New tab control style added

rivisic:
We added new minimalTab format to Pivot control


mpisacic:
if i'm not mistaken, on mockup pivot content was surrounded by border, we could then simply do it here by removing border from pivot, and placing it on pivot item class

rivisic:
@mpisacic it is not a bad idea ;) Done! :)

PR #268, opened 2018-03-23
Filtering fixes for grids

mpisacic:


src/components/QuickGrid/QuickGrid.Props.ts
```diff
@@ -1,6 +1,11 @@
 import * as React from 'react';
 import { ITooltipProps } from '../Tooltip/Tooltip.props';
 
+export interface IQuickGrid {
```
vinkobedek: why is it showing this here? this was a part of some other pull request?

src/components/QuickGrid/DataSelectors.ts
```diff
@@ -64,19 +75,19 @@ const sortRows = (rows: Array<any>, sortColumnName: string,
...
 
-    let members = columns
-        .map(col => col.valueMember);
-    const newRows = rows.filter(row => {      
+    let members = columns.map(col => col.valueMember);
+
+    const newRows = rows.filter(row => {
         let visible = false;
         for (let value of members) {
-            if (row[value].toString().toLowerCase().includes(searchText)) {
```
sglogovic: Why did you replace includes with indexOf?

src/components/QuickGrid/QuickGrid.Props.ts
mpisacic: yes i moved it in other file

src/components/QuickGrid/DataSelectors.ts
mpisacic: indexOf is just older function, its just so that both of grids have the same function

PR #278, opened 2018-04-11
Feat/enable group selection toggle virtualized tree view

josipGrgic11235:
A new prop (isGroupSelectableOnMultiSelect) was added. Its purpose is enabling the user to check a parent item without automatically selecting its child items. This feature is disabled by default. I needed this functionality for my MegaFilter, hope it turns out to be useful somewhere else too.

The select all button functionality isn't quite done yet. The problem is only when there is a search string provided.


josipGrgic11235:
Merging

PR #280, opened 2018-04-17
Feat/grid bool column type

hrestakd:
Added support for boolean DataType column in both QuickGrid and TreeGrid components. Default display is textual. 
To render with icons pick an appropriate BoolFormatType for the optional gridColumn of the same naming.

### DIFF
--- a/src/components/QuickGrid/QuickGrid.Props.ts
+++ b/src/components/QuickGrid/QuickGrid.Props.ts
@@ -4,7 +4,6 @@ import { ITooltipProps } from '../Tooltip/Tooltip.props';
 export interface IQuickGrid {
     scrollToRow(index: number): void;
     updateColumnWidth(columnIndex: number, getWidth: (oldWidth: number) => number): void;
-    formatBoolCell(column: GridColumn, cellData: any): any;
 }
 
 export enum SortDirection {
--- a/src/components/QuickGrid/QuickGrid.Props.ts
+++ b/src/components/QuickGrid/QuickGrid.Props.ts
@@ -99,7 +99,7 @@ export enum BoolFormatTypeEnum {
     /**
      * Checkmark for true, cross for false
      */
-    Both
+    CheckmarkAndCross
 }
 
 export interface GridColumn {
--- a/src/components/QuickGrid/QuickGrid.Props.ts
+++ b/src/components/QuickGrid/QuickGrid.Props.ts
@@ -4,6 +4,7 @@ import { ITooltipProps } from '../Tooltip/Tooltip.props';
 export interface IQuickGrid {
     scrollToRow(index: number): void;
     updateColumnWidth(columnIndex: number, getWidth: (oldWidth: number) => number): void;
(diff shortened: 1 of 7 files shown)


src/components/QuickGrid/QuickGrid.Props.ts
```diff
@@ -4,6 +4,7 @@ import { ITooltipProps } from '../Tooltip/Tooltip.props';
 export interface IQuickGrid {
     scrollToRow(index: number): void;
     updateColumnWidth(columnIndex: number, getWidth: (oldWidth: number) => number): void;
+    formatBoolCell(column: GridColumn, cellData: any): any;
```
vinkobedek: is this to get a custom text value?

vinkobedek: if so, we could move this in the gridcolumn and use it for all data types, not just boolean.

hrestakd: Changed the function so it can be referenced via the existing cellFormatter. Although duplicating the code was now neccessary

src/components/TreeGrid/TreeGrid.tsx
```diff
@@ -77,6 +77,9 @@ export class TreeGrid extends React.PureComponent<ITreeGridProps, ITreeGridState
         };
         expandedColumns.push(replacementFirstColumn);
         for (let i = 1; i < columns.length; i++) {
+            if (columns[i].cellFormatter == null && columns[i].dataType === DataTypeEnum.Boolean) {
```
vinkobedek: won't this change the original columns? do we want that?

src/components/TreeGrid/TreeGrid.tsx
```diff
@@ -260,6 +264,29 @@ export class TreeGrid extends React.PureComponent<ITreeGridProps, ITreeGridState
         );
     }
 
+    private formatBoolCell(cellData: any, rowData: any) {
```
vinkobedek: copy paste from grid? Can we avoid it somehow?

src/components/TreeGrid/TreeGrid.tsx
hrestakd: You're right, this would cause rerenders. Spreading the columns to a new object now

src/components/TreeGrid/TreeGrid.tsx
hrestakd: In the first commit it was possible because the method was exposed by QuickGrid interface, but since we are now setting it before mounting of component, that reference is null

hrestakd: It can be fixed by moving _getTreeColumnsToDisplay to componentDidMount. Unless theres a reason for it to be in the constructor

vinkobedek: you could create a boolFormatterFactory(type: BoolFormatType) to capture the type and thus avoid the need for this and _ref.

hrestakd: Done!

PR #285, opened 2018-04-24
added getObjectValue helper method

hrestakd:
TreeGrid and QuickGrid components now support getting cell values from nested objects.

For this scenario the proper format type of ValueMember or DataMember is `object.property`

### DIFF
--- a/src/components/TreeGrid/TreeGrid.tsx
+++ b/src/components/TreeGrid/TreeGrid.tsx
@@ -10,6 +10,7 @@ import { SpinnerType } from '../Spinner/Spinner.Props';
 import { CellElement } from './CellElement';
 import { ITreeGridProps, ITreeGridState } from './TreeGrid.Props';
 import { boolFormatterFactory } from '../QuickGrid/CellFormatters';
+import { getObjectValue } from '../../utilities/getObjectValue';
 
 export class TreeGrid extends React.PureComponent<ITreeGridProps, ITreeGridState> {
     public static defaultProps = {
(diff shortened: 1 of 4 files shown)


src/components/TreeGrid/TreeGrid.tsx
```diff
@@ -266,30 +267,6 @@ export class TreeGrid extends React.PureComponent<ITreeGridProps, ITreeGridState
         );
     }
 
-    private formatBoolCell(cellData: any, rowData: any) {
```
vinkobedek: what happened here? leftover from a previous move?

src/utilities/getObjectValue.ts
```diff
@@ -0,0 +1,19 @@
+ /**
+ * Get value for object from path.
+ * @param {string} path - Property name of object. For nested objects use property.nested
+ */
+export function getObjectValue(theObject: any, path: string, separator = '.') {
```
vinkobedek: how's the performance of this function? try it with 400 000 elements vs 400 000 elements in the old version just to be on the safe side.

src/components/TreeGrid/TreeGrid.tsx
hrestakd: Yes, it's a sneaky delete, I forgot to clear this last time :)

src/utilities/getObjectValue.ts
hrestakd: I've done some testing, both components on 400-500k rows work almost equally fast

PR #294, opened 2018-05-08
Directional hint prop added

josipGrgic11235:
Added prop so we can set the DirectionalHint of the callout to a value we see fit.


src/components/Dropdown/Dropdown.tsx
```diff
@@ -168,7 +168,9 @@ export class Dropdown extends React.PureComponent<IDropdownProps, IDropdownState
                             gapSpace={0}
                             doNotLayer={false}
                             targetElement={this._dropDown}
-                            directionalHint={DirectionalHint.bottomLeftEdge}
+                            directionalHint={this.props.calloutDirectionalHint ?
```
hbagarich: It would be better to use default props instead of this check in the render method.

PR #302, opened 2018-05-18
Feat/tree grid multiselect

mpisacic:
_Implementation of multiselect on tree grid._
===================================================================
Implementation of multiselect on grid is separated from presentational component (TreeGrid) so that all selection is tracked through **TreeDataSource** object. Correspondence between two components is handled using observers, external subscribers can subscribe to selection and structure changes on data source. When inner collection is changed, TreeDataSource notifies its subscribers that something was changed. What TreeGrid can receive as props is bool that specifies if grid is multi selectable and TreeDataSource which data grid displays.

_There are 3 types of listener that can be registered:_

- **Render listeners** -> React components on which data source calls forceUpdate (we know that something did change, so we can skip shouldComponentUpdate with ease)

- **Selected ids listeners** -> object that wants to receive selected ids on selection change

- **Selected structure listeners** -> objects that wants to receive selected objects on selection change


src/models/TreeData.ts
```diff
@@ -123,11 +417,16 @@ export class TreeDataSource {
...
+                    this.setSelected(existingNode);
+                }
             }
             this.isEmpty = this.treeStructure.children.length === 0;
-            return new TreeDataSource(this);
         }
-        return this;
+        this.notifyWithSelectedIds();
+        this.notifyWithSelectedNodes();
+        this.notify();
```
vinkobedek: this cannot work, without a new reference on tree change we are breaking the redux principle of immutability. Yes it was already broken. but the treeDataSource in combination with the TreeGrid acted as a sort of wrapper around this mutability. With these functions that return a new instance we can write code in our reducers as tate.dataSource=state.dataSource.updateNode(xyz) and i would really, really like it to stay this way. DO NOT merge this in master until the behavior is the same as before ragarding this.

examples/src/TreeGrid.tsx
```diff
@@ -119,8 +162,8 @@ export class Index extends React.Component<any, any> {
                 newChildNode.isExpanded = false;
                 children.push(newChildNode);
             }
-            let newData = this.state.data.updateNode(node.nodeId, { children });
-            this.setState(prev => ({ data: newData }));
+
+            this.state.data.updateNode(node.nodeId, { children });
```
vinkobedek: how does this even work :D. if the state is not changed. what fires the update of the component?

src/components/TreeGrid/TreeGrid.tsx
```diff
@@ -152,21 +166,75 @@ export class TreeGrid extends React.PureComponent<ITreeGridProps, ITreeGridState
...
+                />
+            );
         }
+
+        const onChange = (e) => {
+            e.preventDefault();
+            e.stopPropagation();
+
+            if (this.props.treeDataSource.SelectedNodes[nodeId]) {
+                this.props.treeDataSource.removeSelected(rowData);
```
vinkobedek: removeSelected and setSelected could work like the update function in TreeDataSource  before it was changed. Then you could have normal redux actions and call these functions in the reducer. If you follow the same pattern as it was with the previous version of the update function, the redux store will detect the change automatically.

vinkobedek: actually, disregard this, this is more internal to the treegird. You are correct that this should probably work more along the lines of isExpanded(which also just changes the state directly).

PR #315, opened 2018-06-04
VirtualizedTreeView fix for search + select all

marijahlobik:
We had problem in this scenario:
- we search item in list 
- then click checkbox to select all
- then click checkbox to deselect all => show full list and ignore search

I fixed it in way that onSearch() method return searched list, also I added property in state searchedItemList so I can compare in ComponentWillRecieveProps which list we need to show.

### DIFF
--- a/src/components/TreeFilter/VirtualizedTreeView.tsx
+++ b/src/components/TreeFilter/VirtualizedTreeView.tsx
@@ -48,14 +48,15 @@ export class VirtualizedTreeView extends React.PureComponent<IVirtualizedTreeVie
 
     public static defaultProps = defaultTreeProps;
 
-    public constructor (props: IVirtualizedTreeViewProps) {
+    public constructor(props: IVirtualizedTreeViewProps) {
         super(props);
 
         this.state = {
             partiallyCheckedItemIds: [],
             searchText: props.searchQuery,
             filteredItems: ItemOperator.filterItems(props.items, props.searchQuery),
-            scrollToIndex: undefined
+            scrollToIndex: undefined,
+            searchedItemList: null
         };
 
         const lookups = this.props.lookupTableGetter(props.items);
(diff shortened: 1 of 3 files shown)


src/components/TreeFilter/VirtualizedTreeView.tsx
```diff
@@ -77,21 +78,41 @@ export class VirtualizedTreeView extends React.PureComponent<IVirtualizedTreeVie
...
+        const newItems = [];
+        if (this.state.searchedItemList !== null) {
+            this.state.searchedItemList.map((data) => {
+                const item = nextProps.items.find(x => x.id === data.id);
+                if (item) {
+                    newItems.push(item);
+                }
+            });
+
+            if (newItems.length !== 0 && newItems !== this.props.items) {
```
vinkobedek: won't this always be true? newItems = [] on every props receive.

src/components/TreeFilter/VirtualizedTreeView.tsx
```diff
@@ -379,14 +400,16 @@ export class VirtualizedTreeView extends React.PureComponent<IVirtualizedTreeVie
         this.setState(prevState => ({
             ...prevState,
             searchText: searchText,
-            filteredItems: newItems
+            filteredItems: newItems,
```
vinkobedek: what is the difference between filteredItems and searchedItemList

src/components/TreeFilter/VirtualizedTreeView.tsx
```diff
@@ -77,21 +78,41 @@ export class VirtualizedTreeView extends React.PureComponent<IVirtualizedTreeVie
...
+        if (this.state.searchedItemList !== null) {
+            this.state.searchedItemList.map((data) => {
+                const item = nextProps.items.find(x => x.id === data.id);
+                if (item) {
+                    newItems.push(item);
+                }
+            });
+
+            if (newItems.length !== 0 && newItems !== this.props.items) {
+                const filteredItems = ItemOperator.filterItems(newItems, this.state.searchText);
```
vinkobedek: why can't we just do` const filteredItems = ItemOperator.filterItems(nextProps.items` as it was before?

src/components/TreeFilter/VirtualizedTreeView.tsx
marijahlobik: Only in case that we have searchedItems it will push those items in newItems. This:
if (newItems.length !== 0 && newItems !== this.props.items) 
first check if we have items in list and if we have then length will be !== 0, if it is just initialized length is 0. 

It's important to keep searched items in newList because on every change of checkbox component receive new props and we don't lose searched items unless we clear search field. It can happen that you searched items -> select all -> deselect all because you maybe want just one of those searched -> and you lose searched item but you still have searched text in field.

src/components/TreeFilter/VirtualizedTreeView.tsx
marijahlobik: Because when we search something we want only searched items until we clear search field and nextProps.items is list of all items.

src/components/TreeFilter/VirtualizedTreeView.tsx
marijahlobik: Well, searchedItemList as it name says is list of searcched items. 

I spent quite some time to figure out what filtered items property is and how it works. I couldn't find why and where it set filteredItems as list of all items when we have active search field and deselect all items so I added new property that with intiutive name tell what it is.

src/components/TreeFilter/VirtualizedTreeView.tsx
vinkobedek: always as in always in this code block. It will still cause the logic to be triggered each and every time there is something in the searchedItemList.

But i am still puzzled by newItems, searchedItems, filteredItems. What is what and why? 
Why do i need to filter based on newItems instead of props.items? visible items depend only on the text entered.
Why are newItems a parameter in the onItemsSearch callback? are they just put into props.searchedItemList?

marijahlobik: > It will still cause the logic to be triggered each and every time there is something in the searchedItemList.

Yes, and I think we need it because of the problem that I mentioned. 

> But i am still puzzled by newItems, searchedItems, filteredItems. What is what and why?

I'm going to check if I can somehow refactor it to be less confusing

> Why are newItems a parameter in the onItemsSearch callback?  

When we are calling component VirtualizedTreeView from our project we need to send filterSelecition (which means that we need to have enum (All = 0, None = 1,  Selected = 2)) and I needed it because of that but I think that I can delete it now if I'm checking in ComponentWillRecieveProps. I'll test it.

PR #321, opened 2018-06-14
Bugfix on left navigation selected item display

mpisacic:
There was mixup between inner state of component and what was source of truth from props. Removed state that memorizes current selected item so now whole component is controlled.


hbagarich:
It seem as if this pull request can introduce some breaking changes in the way the Navigation component is working? 
Can we verify that other products are working after this change?

mpisacic:
It should be checked in SSM is everything is still working properly.
In SysKit Insights new changes are working as it should.

PR #324, opened 2018-07-02
Feat/pivot icon with text support

hrestakd:
Pivot items now support icons with text in title. Behaviour is decided via the new optional PivotItem prop linkRenderMode.
If not provided, rendering works same as before.


src/components/Pivot/PivotItem.Props.ts
```diff
@@ -1,8 +1,14 @@
 import * as React from 'react';
 
+export enum renderModeEnum {
```
vinkobedek: how come little r in _renderModeEnum_

hrestakd: typo

hrestakd: fixed :)

src/components/Pivot/Pivot.tsx
```diff
@@ -117,17 +112,37 @@ export class Pivot extends React.Component<IPivotProps, IPivotState> {
...
     }
 
+    private _linkRenderer(link: IPivotItemProps): JSX.Element {
+        const { textSize } = this.props;
+        let textStyle;
+        if (textSize) {
+            textStyle = {fontSize: textSize};
+        }
+        const textElement = <span className={'pivot-text'} style={textStyle}>{link.linkText}</span>;
+        const iconElement = <Icon iconName={link.linkIcon} className={'pivot-icon'} title={link.linkText} />;
```
vinkobedek: it bugs me a little that an icon element is created if i do not specify a linkIcon. But i will not hold it againts this PR.

PR #327, opened 2018-07-12
disabled nodes functionality

vinkobedek:
I needed to disable some nodes that were shown in the TreeGrid. Not sure if i should maybe put isNodeDisabled inside $meta. If i do, it is cleaner and it cannot be used by mistake(i.e. there is a property in the C# classes that is named nodedisabled, what if i did not remember this fact). If i do put it there i can't just use it without additional logic.

### DIFF
--- a/src/components/TreeGrid/TreeGrid.tsx
+++ b/src/components/TreeGrid/TreeGrid.tsx
@@ -225,7 +225,7 @@ export class TreeGrid extends React.PureComponent<ITreeGridProps, ITreeGridState
         };
 
         let checkBox: JSX.Element = null;
-        if (this.props.isMultiSelectable) {
+        if (this.props.isMultiSelectable && !rowData.isNodeDisabled) {
             const checked = this.props.treeDataSource.SelectedNodes.hasOwnProperty(nodeId);
             checkBox = (
                 <VirtualizedTreeViewCheckBox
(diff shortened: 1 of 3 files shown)


src/components/TreeGrid/TreeGrid.tsx
```diff
@@ -278,8 +282,11 @@ export class TreeGrid extends React.PureComponent<ITreeGridProps, ITreeGridState
             rowClass,
             column.cellClassName,
             rowData.className,
-            { 'border-column-cell': notLastIndex },
-            { 'is-selected': isSelectedRow && this.props.highlightRowsInMultiSelect });
+            {
+                'border-column-cell': notLastIndex,
+                'is-selected': isSelectedRow && this.props.highlightRowsInMultiSelect && !rowData.isNodeDisabled,
```
kacavendadino: @vinkobedek does this mean i cannot have a selected node that is also disalbed? (example for a manage form that shows existing connections, but they cannot be removed(unchecked))?

vinkobedek: i didnt want to spend to much time on the multiselect logic so i left it as is after graying out the rows and hiding the checkboxes. the checkbox will not be shown and you will not be able to select it, but if you select the parent node of the selected node that it will be added to the selected list regardless of the fact that it is disabled, or if you select all of the child nodes of a disabled parent. Also, the nodes are still selectable programmatically by using the datasource. 

To be more flexible we should add some more properties to define the selection behavior better but the current behavior surves my requirements where i need to show a list of tenant to select for autodiscovery but some of them are not autodiscoverable.

kacavendadino: makes sense

PR #337, opened 2018-08-07
This is supposed to enable Nested objects to be used as grid data

Cubelaster:
Basically added a bunch of .split('.') and foreach inside of existing methods for handling sorts/groupings and other grid manipulation methods. 
Don't know if there are more places that this needs to be added. 
I tried with 200k rows and it showed no delay.

### DIFF
--- a/src/components/QuickGrid/DataSelectors.ts
+++ b/src/components/QuickGrid/DataSelectors.ts
@@ -79,22 +79,24 @@ const filterRows = (rows: Array<any>, columns: Array<GridColumn>, searchText: st
     if (!searchText) {
         return rows;
     }
-
     let filterText = searchText.toLowerCase();
-
     let members = columns.map(col => col.valueMember);
-
     const newRows = rows.filter(row => {
         let visible = false;
         for (let value of members) {
-            if (row[value].toString().toLowerCase().indexOf(filterText) !== -1) {
-                visible = true;
-                break;
+            let objectProps = value.split('.');
+            let result = row;
+            for (let i = 0; i < objectProps.length; i++) {
+                result = result[objectProps[i]];
+                if (result != null && result.toString()
+                    .toLowerCase().indexOf(filterText) !== -1) {
+                    visible = true;
+                    break;
+                }
             }
         }
         return visible;
     });
-
     return newRows;
 };
 
--- a/src/components/QuickGrid/DataSelectors.ts
+++ b/src/components/QuickGrid/DataSelectors.ts
@@ -1,7 +1,7 @@
 import * as _ from 'lodash';
 import { createSelector } from 'reselect';
 
-import { sortArray } from '../../utilities/array';
+import { sortRowsByColumn } from '../../utilities/array';
 import {
     DataTypeEnum,
     GridColumn,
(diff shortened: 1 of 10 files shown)


src/utilities/array.ts
```diff
@@ -56,7 +63,22 @@ export const sortArray = (inputArray: Array<any>, sortOptions: Array<SortProps>)
 
 export const groupBy = function (inputArray, groupProp) {
     return inputArray.reduce((groups, item) => {
-        (groups[item[groupProp]] = groups[item[groupProp]] || []).push(item);
+        let result = item;
+        let splits = groupProp.split('.');
+        for (let i = 0; i < splits.length; i++) {
```
vinkobedek: won't this for loop just set the result to the deepest value?

src/utilities/array.ts
```diff
@@ -38,8 +38,15 @@ export const sortArray = (inputArray: Array<any>, sortOptions: Array<SortProps>)
                 valueA = sortOption.sortFunction(a);
                 valueB = sortOption.sortFunction(b);
             } else {
-                valueA = a[sortOption.column];
-                valueB = b[sortOption.column];
+                let splits = sortOption.column.split('.');
+                let resultA = a;
+                let resultB = b;
+                for (let i = 0; i < splits.length; i++) {
```
vinkobedek: same as for the groupinng operation, won't this just set valueA and ValueB to the deepest value?

src/components/QuickGrid/DataSelectors.ts
```diff
@@ -79,22 +79,24 @@ const filterRows = (rows: Array<any>, columns: Array<GridColumn>, searchText: st
...
-
     let members = columns.map(col => col.valueMember);
-
     const newRows = rows.filter(row => {
         let visible = false;
         for (let value of members) {
-            if (row[value].toString().toLowerCase().indexOf(filterText) !== -1) {
-                visible = true;
-                break;
+            let objectProps = value.split('.');
```
vinkobedek: how's the performance impact of these changes?

src/utilities/array.ts
Cubelaster: It will. Can it be done better?

src/components/QuickGrid/DataSelectors.ts
Cubelaster: Switched from 100k to 200k and no impact.

src/utilities/array.ts
Cubelaster: Deepest FOUND.

src/utilities/array.ts
Cubelaster: I saw a method written by @hrestakd `getObjectValue` so I will try to use that instead

vinkobedek: ah, sorry, my bad, now that i am looking at this again i see what you are actually doing here. All good.

hrestakd: the method itself was really fast but the sorting for tree grid was the slow part so I removed it

hrestakd: For QuickGrid purposes it has little to no performance impact

Cubelaster:
Now switched to getObjectValue(item, prop);
http://take.ms/E3AUu
Performance video. 
This seems to make a hit on performance, but seems ok?

PR #341, opened 2018-09-03
Notification bubble on icon

lbalasko:
This PR is for notification bubble on icon. You pass the number you wish to show in bubble by props. In case that no number is passed, the code is the same as before. In case of number being set, first a container is created that contains both icon and notification bubble. Size of this container is set based on size of the icon. Notification bubble is set in right top corner of container. Because size of container and notification bubble are both based on size of icon, this positions notification bubbles center on right top corner of icon (with big icons it doesn't look like that because there is a lot of white space in their corners, but it is centered). I uses style object to set css attributes for container, notification bubble and notification number. You can overwrite any of them by passing style object with attribute you wish to change. For attributes that are not set in style object, you can use class names. There are examples on icon and left navigation pages.

### DIFF
--- a/src/components/NotificationIcon/NotificationIcon.tsx
+++ b/src/components/NotificationIcon/NotificationIcon.tsx
@@ -1,8 +1,14 @@
 import * as React from 'react';
 import { INotificationIconProps, DefaultIconNumberStyleObject, NotificationBubbleStyleObject} from './NotificationIcon.Props';
 import { Icon } from '../Icon/Icon';
+import * as classNames from 'classnames';
+import './NotificationIcon.scss';
 
 export const NotificationIcon: (props: INotificationIconProps) => JSX.Element = (props: INotificationIconProps) => {
+
+    let iconClassName = classNames(
+        'icon-with-notification',
+        [props.className]);
 
     const numberString = props.notificationNumber === undefined || props.notificationNumber === 0 ? '' 
                             : props.notificationNumber < 10 ? props.notificationNumber.toLocaleString() : '9+';
@@ -25,7 +31,7 @@ export const NotificationIcon: (props: INotificationIconProps) => JSX.Element = 
 
     return (
     <div style={styleObject.containerStyleObject as React.CSSProperties} className={props.containerClassName}>
-        <Icon iconName={props.iconName} className={props.className} iconSize={props.iconSize}
+        <Icon iconName={props.iconName} className={iconClassName} iconSize={props.iconSize}
                 width={props.width} height={props.height}></Icon>
         {(numberString !== '') &&
             <div style={styleObject.bubbleStyleObject as React.CSSProperties} className={props.notificationBubbleClassName}>
--- a/src/components/NotificationIcon/NotificationIcon.Props.ts
+++ b/src/components/NotificationIcon/NotificationIcon.Props.ts
@@ -21,11 +21,10 @@ export interface NotificationBubbleStyleObject {
 }
 
 export interface ContainerStyleObject {
-  width?: string;
-  height?: string;
   position?: string;
   display?: string;
-  alignItems?: string;
+  marginRight?: string;
+  marginTop?: string;
 }
 
 export interface BubbleStyleObject {
@@ -36,6 +35,7 @@ export interface BubbleStyleObject {
   height?: string;
   borderRadius?: string;
   backgroundColor?: string;
+  transform?: string;
 }
 
 export interface NumberStyleObject {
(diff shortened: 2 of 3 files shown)


src/components/Icon/Icon.tsx
```diff
@@ -39,17 +41,107 @@ export const Icon: (props: IIconProps) => JSX.Element = (props: IIconProps) => {
         iconWidth = props.width || '16px';
     }
 
+    const numberString = props.notificationNumber === undefined ? '' : props.notificationNumber < 10 ?
+                            props.notificationNumber.toLocaleString() : '9+';
+    const defaultStyleObject = DefaultIconNumberStyleObject(props.iconSize, numberString.length);
+    let styleObject: NotificationBubbleStyleObject = {numberStyleObject: {}, containerStyleObject: {}, bubbleStyleObject: {}};
+
+    if (props.notificationBubbleStyleObject) {
```
vinkobedek: maybe use Object.assign or something to avoid manual setting of the values?

Cubelaster: the point here is the merge option. object.assign will override any default option, while this allows for partial style object, if i'm not wrong

lbalasko: yes, the point is that you can assign only, for example, color to blue and that all other values are default... that's why I need to assign manualy

src/components/LeftNavigation/LeftNavigation.Props.ts
```diff
@@ -24,6 +24,7 @@ export interface ILeftNavigationOption {
     selected?: boolean;
     disabled?: boolean;
     position?: LeftNavigationOptionPositionEnum;
+    notificationNumber?: number;
```
vinkobedek: could this work with ie. any label? say i wanted to add the 'New' text above an icon?

lbalasko: no, the point is that it is a number. this is only bubble for notifications. and this needs to be a number so that I can check if number is bigger then 10, then only 9+ is shown.

src/components/Icon/Icon.tsx
vinkobedek: i'm afraid i do not understand.
is this not the same?

```javascript
 styleObject: NotificationBubbleStyleObject = {
containerStyleObject: {
    ...defaultStyleObject.containerStyleObject
}}

if(props.notificationBubbleStyleObject.containerStyleObject) {
 object.assign(styleObject.containerStyleObject, props.notificationBubbleStyleObject.containerStyleObject)
}
;```

File: src/components/LeftNavigation/LeftNavigation.Props.ts (hunk 2 above)
Comment (hbagarich) [inline src/components/LeftNavigation/LeftNavigation.Props.ts]: I'd like to see a specialized version of Icon, called NotificationIcon or some kind of CompositeIcon which can combine multiple SVGs. That way we can have Icons with "NEW" text or something similar, and also combine two Icons programatically if there is need?
I feel like this notificationNumber is not a part of generic Icon component..

File: src/components/Icon/Icon.tsx (hunk 1 above)
Comment (lbalasko) [reply src/components/Icon/Icon.tsx]: if props.notificationBubbleStyleObject.containerStyleObject has only width set, then when it is asigned to styleObject.containerStyleObject, styleObject.containerStyleObject will also have only width set. what we wanted here is to have styleObject.containerStyleObject with default values for height, position, display and alignItems and only width from props.notificationBubbleStyleObject.containerStyleObject. this way when you wish to set only width, you can, you don't have to set height, position, display and alignItems too.

Comment (vinkobedek) [reply src/components/Icon/Icon.tsx]: I dont get it.
that is why i used the spread operator when initialiing containerStyleObject, to set the default values. Object.assign will the use the partial object to override the rest of them.

Object.assign({width:10, height:10}, {widht:12}) will produce {widht:12, height: 10}

true, Object.assign({width:10, height:10}, {widht:null}) will produce {widht:null, height: 10}, but i don't see this as a problem.

Comment (vinkobedek) [reply src/components/Icon/Icon.tsx]: unless you specifically wanted to support this case.

Comment (lbalasko) [reply src/components/Icon/Icon.tsx]: you are right. I apologize, I didn't realize it right away. I will change it. Thank you.

File: src/components/LeftNavigation/LeftNavigation.Props.ts (hunk 2 above)
Comment (lbalasko) [reply src/components/LeftNavigation/LeftNavigation.Props.ts]: I removed NotificationIcon from Icon component. NotificationIcon is now a separate component.

File: src/components/NotificationIcon/NotificationIcon.tsx [hunk 3]
```diff
@@ -0,0 +1,92 @@
...
+        default:
+        iconHeight = props.height || '16px';
+        iconWidth = props.width || '16px';
+    }
+
+    const numberString = props.notificationNumber === undefined || props.notificationNumber === 0 ? '' 
+                            : props.notificationNumber < 10 ? props.notificationNumber.toLocaleString() : '9+';
+    const defaultStyleObject = DefaultIconNumberStyleObject(props.iconSize, numberString.length);
+   
+    let styleObject: NotificationBubbleStyleObject = { ...defaultStyleObject};
```
vinkobedek: down below on line 49. will actually also modify the defaultStyleObject since the containerStyleObject of styleObject is te same reference if you think about it. so if we are doing it like this then the spread operator is not even necessary since the defaultStyleObject will be mutated anyways. So there are two choices.
1. rename the DefaultIconNumberStyleObject function to CreateDefaultIconNumberStyleObject so that it is evident that mutation was intended
2. spread all of the subobjects to a different reference will be used below when using object.assign

PR #356, opened 2018-10-24
Render on select all forced.

MatijaMaric:
- fixed not re-rendering TreeView component on select all checkbox toggle
- filterSelection is delegated to List component, if it's changed the component will re-render

### DIFF
--- a/src/components/TreeFilter/VirtualizedTreeView.tsx
+++ b/src/components/TreeFilter/VirtualizedTreeView.tsx
@@ -125,9 +125,6 @@ export class VirtualizedTreeView extends React.PureComponent<IVirtualizedTreeVie
             }
         } else if (!_.isEqual(this.props.filterSelection.selectedIDs, prevProps.filterSelection.selectedIDs)) {
             selectionChanged = true;
-            if (this._list != null) {
-                this._list.forceUpdateGrid();
-            }
         }
         if (selectionChanged) {
             if (this.props.filterSelection.type !== FilterSelectionEnum.All) {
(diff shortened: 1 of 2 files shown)


src/components/TreeFilter/VirtualizedTreeView.tsx
```diff
@@ -642,6 +642,7 @@ export class VirtualizedTreeView extends React.PureComponent<IVirtualizedTreeVie
             this.props.onValuesSelected(this.props.filterId, { type: FilterSelectionEnum.None, selectedIDs: [] });
         }
         this.setState(prevState => ({ ...prevState, partiallyCheckedItemIds: newPartiallyChecked }));
+        this._list.forceUpdateGrid();
```
vinkobedek: how come this state change didn't trigger an update with the regular lifecycle?

MatijaMaric: state only stores info on partially checked list. if none/all of it is selected, state doesn't change.

src/components/TreeFilter/VirtualizedTreeView.tsx
```diff
@@ -181,6 +181,7 @@ export class VirtualizedTreeView extends React.PureComponent<IVirtualizedTreeVie
                             rowRenderer={this.rowRenderer}
                             rowCount={this.state.filteredItems.length}
                             scrollToIndex={this.state.scrollToIndex}
+                            {...this.props.filterSelection}
```
vinkobedek: is the forceUpdate in componentDidUpdate still relevant? or can it be removed now?

MatijaMaric: yeah, i think it can go now

PR #363, opened 2018-11-05
Added column name to custom grouped row values

zlatkoHrastic:
Added column name to custom grouped row values in QuickGrid


src/components/QuickGrid/QuickGrid.Props.ts
```diff
@@ -49,7 +49,7 @@ export interface IQuickGridProps {
     customRowSorter?: (sortBy, sortDirection) => void;
     customCellRenderer?: (args: ICustomCellRendererArgs) => React.ReactNode;
     onGroupByChanged?: (groupBy: Array<IGroupBy>) => void;
-    groupRowFormat?: (rowData: any) => string;
+    groupRowFormat?: (rowData: any, columnName?: string) => string;
```
kacavendadino: Maybe change return type to React.Node so we have greater flexibility (it shouldn't break existing use cases)

zlatkoHrastic: Fixed and tested for backward compatibility

PR #367, opened 2018-11-20
Feat/icon with number

lbalasko:
NotificationBubbleStyleObject was removed from ILeftNavigationProps and added to ILeftNavigationOptions so that every option can have different style object for notification number.


src/components/LeftNavigation/LeftNavigation.Props.ts
```diff
@@ -29,6 +28,7 @@ export interface ILeftNavigationOption {
     disabled?: boolean;
     position?: LeftNavigationOptionPositionEnum;
     notificationNumber?: number;
+    notificationBubbleStyleObject?: NotificationBubbleStyleObject;
```
vinkobedek: why not both? set one default style for the menu and be able to override it per menu item?

lbalasko: Added both options.

PR #373, opened 2018-11-26
do not update selected node id on multi select

hrestakd:


src/components/TreeGrid/TreeGrid.tsx
```diff
@@ -128,7 +126,7 @@ export class TreeGrid extends React.PureComponent<ITreeGridProps, ITreeGridState
         this._quickGrid.updateColumnWidth(1, (old) => {
             return Math.max(old, getColumnMinWidth(this.state.columnsToDisplay[1]));
         });
-        if (this.props.selectedNodeId !== nextProps.selectedNodeId && this.props.isNodeSelectable) {
+        if (!nextProps.isMultiSelectable && (this.props.selectedNodeId !== nextProps.selectedNodeId || nextProps.isMultiSelectable !== this.props.isMultiSelectable) && this.props.isNodeSelectable) {
```
vinkobedek: something seems funny here. if the second part of the condition is dependant on nextProps.isMultiSelectable being false would't it be the same as writing this.props.isMultiSelectable = true? Then it would read `if we are not multiselectable and the selectedNodeId changed or we were multiselectable fire the selectedEvent`.

hrestakd: that's right, shortened now

PR #381, opened 2018-12-18
QuickGrid column picker

MatijaMaric:
- [x] Added a new component for choosing columns to show/hide in grid.
- [x] Optional dropdown button in grid header.
- [x] Live rerender and readjusting column widths to fit total width of grid
- [ ] Support for TreeGrid ?

screen

### DIFF
--- a/src/components/QuickGrid/QuickGrid.Props.ts
+++ b/src/components/QuickGrid/QuickGrid.Props.ts
@@ -176,7 +176,7 @@ export interface QuickGridActions {
 export interface ContextActionsObject {
     actions: Array<ActionItem>;
     dropdownIconName?: string;
-    dropdownCustomRenderer?: (rowIndex: number, actions: Array<ActionItem>, onActionClicked: (rowIndex: number, action: ActionItem) => void) => JSX.Element;
+    dropdownCustomRenderer?: (rowIndex: number, actions: Array<ActionItem>, onMenuToggle: (opened: boolean) => void, onActionClicked: (rowIndex: number, action: ActionItem) => void) => JSX.Element;
 }
 
 export function isContextActionsObject(arg: any): arg is ContextActionsObject {
--- a/src/components/QuickGrid/QuickGridRowContextActionsHandler.tsx
+++ b/src/components/QuickGrid/QuickGridRowContextActionsHandler.tsx
@@ -151,7 +151,7 @@ export class QuickGridRowContextActionsHandler extends React.PureComponent<IQuic
                     onClosed={() => onMenuToggle(false)}
                     delayMs={this.props.delayMs}
                 >
-                    {actions.dropdownCustomRenderer(rowIndex, dropdownActions, onActionClicked)}
+                    {actions.dropdownCustomRenderer(rowIndex, dropdownActions, onMenuToggle, onActionClicked)}
                 </Dropdown>
             );
         } else if (dropdownActions && dropdownActions.length > 0) {


src/components/QuickGrid/ColumnPicker/ColumnPicker.tsx
```diff
@@ -0,0 +1,89 @@
...
+    public constructor(props: ColumnPickerProps) {
+        super(props);
+        this.state = {
+            isColumnSelected: this.setVisibleColumns(props.columns, props.pickedColumns)
+        };
+    }
+
+    private setVisibleColumns(columns: Array<GridColumn>, picked?: Array<GridColumn>): {[key: string]: boolean} {
+        let visible: {[key: string]: boolean} = {};
+        columns.forEach(col => visible[col.valueMember] = !!!picked);
```
vinkobedek: why the triple !!! instead of a single?

MatijaMaric: It's an optional field, so it's first converted to boolean `!!(undefined)` and then inverted

src/components/QuickGrid/ColumnPicker/ColumnPicker.tsx
```diff
@@ -0,0 +1,89 @@
...
+        if (picked) {
```
kacavendadino: why do we have to iterate through picked columns again? why don't we check it when iterating through all columns (should picked be a subset of columns)? visible[col.valueMember] = !!!picked[col.valueMember]

src/components/QuickGrid/QuickGridHeader.Props.ts
```diff
@@ -20,6 +20,9 @@ export interface IGridHeaderProps {
     onCollapseAll?: (event) => void;
     onExpandAll?: (event) => void;
     tooltipsEnabled?: boolean;
+    hasColumnPicker?: boolean;
+    pickedColumns?: Array<GridColumn>;
+    onColumnsPicked?(picked: Array<GridColumn>);
```
vinkobedek: maybe onColumnSelectionChanged?

src/components/QuickGrid/ColumnPicker/ColumnPicker.tsx
```diff
@@ -0,0 +1,89 @@
...
+export default class ColumnPicker extends React.Component<ColumnPickerProps, ColumnPickerState> {
+
+    public constructor(props: ColumnPickerProps) {
+        super(props);
+        this.state = {
+            isColumnSelected: this.setVisibleColumns(props.columns, props.pickedColumns)
+        };
+    }
+
+    private setVisibleColumns(columns: Array<GridColumn>, picked?: Array<GridColumn>): {[key: string]: boolean} {
```
kacavendadino: isn't this set selected columns not visible?

src/components/QuickGrid/ColumnPicker/ColumnPicker.tsx
MatijaMaric: It wouldn't be any more optimized, `picked` is an array of objects, not a dictionary, so I would have to do a lookup for each column.
This way I only do two passes, and not even that because `picked` is a subset of all columns

src/components/QuickGrid/ColumnPicker/ColumnPicker.tsx
MatijaMaric: it's just a helper function to set values of checkboxes

MatijaMaric: yeah actually you're right 😄

src/components/QuickGrid/ColumnPicker/ColumnPicker.tsx
MatijaMaric: It's working with a single !. This is not the first time I used this operator :joy:

PR #390, opened 2019-01-10
Fix for grid crashing without pickedColumns prop.

MatijaMaric:


src/components/QuickGrid/QuickGrid.tsx
```diff
@@ -194,7 +194,7 @@ export class QuickGridInner extends React.Component<IQuickGridProps, IQuickGridS
             });
         }
 
-        const _pickedColumns = pickedColumns || this.state.pickedColumns;
+        const _pickedColumns = pickedColumns || (this.state && this.state.pickedColumns);
```
kacavendadino: isnt it better to transform this function to an arrow function so state will always be correctly bonded? if anyone ever tries to use state here again, it will crash again (props too)

MatijaMaric: yeah, i see what you mean. that makes more sense

PR #392, opened 2019-01-22
Fix/case insensitive sorting

MatijaMaric:
- used Intl.Collator for minimal performance impact (https://jsperf.com/sort-locale-strings/5)
- groupBy rows now correctly use same methods to resolve value as cells

### DIFF
--- a/src/components/TreeGrid/TreeGridDataSelectors.ts
+++ b/src/components/TreeGrid/TreeGridDataSelectors.ts
@@ -120,6 +120,10 @@ const sort = (input, sortDirection, sortColumn, valueGetterForSort) => {
             valueA = a[sortColumnFinal];
             valueB = b[sortColumnFinal];
         }
+        if (typeof valueA === 'string' && typeof valueB === 'string') {
+            valueA = valueA.toLowerCase();
+            valueB = valueB.toLowerCase();
+        }
         if (valueA < valueB) {
             return -1 * sortModifier;
         }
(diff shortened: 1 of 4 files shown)


src/components/TreeGrid/TreeGridDataSelectors.ts
```diff
@@ -108,34 +108,32 @@ const sort = (input, sortDirection, sortColumn, valueGetterForSort) => {
         return input;
     }
     const sortModifier = sortDirection === SortDirection.Descending ? -1 : 1;
+    const collator = Intl.Collator([...navigator.languages], {sensitivity: 'accent', numeric: true});
```
vinkobedek: will this dramatically slow down the sort process by any chance?

src/components/TreeGrid/TreeGridDataSelectors.ts
```diff
@@ -98,7 +98,7 @@ const sortData = (treeNode: AugmentedTreeNode, sortColumn: string, sortDirection
     // if the last sort configuration differs from the current, we need to resort the children
     // otherwise, performance gains
     if ((<AugmentedTreeNode>treeNode).$meta.sortRequestId !== rootSortRequestId) {
-        sort(treeNode.children, sortDirection, sortColumn, valueGetterForSort);
+        treeNode.children = sort(treeNode.children, sortDirection, sortColumn, valueGetterForSort);
```
vinkobedek: before the code deliberately used the same array to sort the nodes to avoid allocating new memory. have you noticed any large memory/cpu differences with this new code?

vinkobedek: try sorting a  couple of times with 500k nodes with the old code and the new one

MatijaMaric: lodash doesn't allow mutable sort, i'll try writing my own implementation

vinkobedek: ok, but no need for complications if you find out that there is no memory/CPU pressure from this.

vinkobedek: ok

PR #396, opened 2019-02-06
added calloutClassName to TreeFilter

gudanmarko:
- added calloutClassName to TreeFilter
- version: 0.14.19


package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "quick-react-ts",
-    "version": "0.14.18",
```
zlatkoHrastic: this should be done in package.json not in here. This is autogenerated file

PR #397, opened 2019-02-12
Feat/slider tick marks

gudanmarko:
- added possibility to turn on tick marks on slider
- version 0.14.20


src/components/Slider/Slider.tsx
```diff
@@ -106,7 +107,8 @@ export class Slider extends CommonComponent<ISliderProps, ISliderState> implemen
                                 style={isRTL() ?
                                     { 'right': thumbOffsetPercent + '%' } :
                                     { 'left': thumbOffsetPercent + '%' }}>
-                            </span>
+                            </span>    
+                            {showTickMarks && this._renderTickMarks(max - min).map(x => x)}
```
zlatkoHrastic: do you need map function in here?

gudanmarko: You were right. It's not needed.

